Issue #60, opened 2015-01-07
Add support for the rest of JetBrains IDEs


Tinche:
Idea Ultimate, PyCharm Commercial Edition, PhpStorm, RubyMine, WebStorm, CLion.

As far as I remember they actually display the license themselves when started for the first time, so no problems there. This should be very easy with the Idea/PyCharm support we already have, assuming the hashes are located in their usual places.

I'll do this myself, one by one, but if anyone else wants to contribute this would be a great opportunity.

syndbg:
I would like to try my luck with adding PyCharm Commercial Edition and RubyMine.

However I'll wait a bit to see where your PR #65 is going. It'll surely help me along the way.

didrocks:
And merged!

If you need any help on where and how to start, feel free to ask! #65 is a great way to start, indeed, as you have large and medium tests into it :)

Tinche:
Hi, @syndbg,

here's what I suggest:

1. implement the actual code
2. test manually that it works
3. write the accompanying large test, check that it passes

Now you just need to write a medium test. These tests are a little more complex since they use a special customized Docker environment. The medium test you want is basically a large test executing against a fake (our) web server, serving fake (our) files. For the medium test, I suggest:

1. scrape the download/thank you page for the product you're working on, save it in the appropriate place in the server-content directory
2. edit it to point to a fake filename
3. create the fake file by using an actual download, then stripping out libraries and stuff from it until it's ~100kb. Then replace the startup script with a similar one from our PyCharm/Idea dummy files. Also add the fake-java.sh script.
4. generate a sha256 checksum into an appropriate file (filename + ".sha256)

That's my workflow for these. Good luck!

syndbg:
Thanks for the heads-up!  I've set up my dev venv and I will try and implement the PR in a day or so.

didrocks:
All but CLion are now in master. CLion is opened for a great contribution, taking the other ones already merged as an example :)

praetore:
I would like to see CLion in there :D

tony:
I see CLion now, but not  CLion Professional yet.

LyzardKing:
There's only one version of CLion. 
The version we install has a 30 day trial license.
Is this what you meant?

PR #65, opened 2015-01-17
Added Idea Ultimate.

Tinche:
Here's IntelliJ Idea Ultimate edition, as asked for in #60.

Open points:

* no medium tests - they would be almost identical to the medium tests for Idea, community edition, and would require upgrading our mock http server to differentiate based on query parameters (the URLs are the same)

* right now the command is 'idea-ultimate', and the community edition is just 'idea'. What should be do for PyCharm and the others - there's no PyCharm ultimate, and RubyMine only has the commercial version? Maybe the commercial versions should be 'idea' and 'pycharm', and the community editions 'idea-community' and 'pycharm-community'?

* Changed the .desktop generation for existing JetBrains slightly to match the .desktop files they themselves generate


didrocks:
Thanks! doing the review in a few minutes. Answering your open points first :)

* I think medium tests can be valuable still to protect us against issues. For instance, with some fixes for android studio, I didn't notice at first that I dropped the "s" of https url. Of course, the large tests were still working, but not the medium ones and I could fix it. Another way it can protect us is if we change without noticing some paths. Of course, large tests will fail, but seeing that the medium were as well failing will show where to start looking at (meaning: our side :)). Ensuring that the medium test pass but not the large ones will show us that something changed on the server/application side, and not ubuntu make.
It's not a lot to bring those tests, but I think it still worths it, what do you think?

* I like that the default is to download and install the community versions and I don't think we should change that, so your current implementation with -ultimate is what I think makes more sense. The description of each IDE you have set is enough I guess. For pycharm, let's just keep the current pycharm and RubyMine, maybe calling it rubymine-ultimate only?

* desktop generation: I looked at the pycharm one. I'm not found of their Comments, maybe can you send a message to them as you have some contacts? That will make matching in the unity dash bad as we look at the comment and it's not self-explanatory. You should add (as it's the case for android studio) the StartupWMClass=jetbrains-pycharm through the extra parameter. The remaining parts are the same, aren't they?

umake/frameworks/ide.py
```diff
@@ -181,7 +181,7 @@ def post_install(self):
         comment = self.description + " (UDTC)"
         categories = "Development;IDE;"
         create_launcher(self.desktop_filename,
-                        get_application_desktop_file(name=self.name,
+                        get_application_desktop_file(name=self.description,
```
didrocks: Why changing the name by description here? I think it's fine to say that we well have idea and idea-ultimate. Any good shell will as well propose the description as a matching pattern.

umake/frameworks/ide.py
```diff
@@ -211,10 +211,26 @@ class Idea(BaseJetBrains):
 
     def __init__(self, category):
         super().__init__(name=_("Idea"),
-                         description=_("IntelliJ Idea Community Edition"),
+                         description=_("IntelliJ IDEA Community Edition"),
```
didrocks: You are going to impact translation. Remind me when merging to change that in the .po file and reupload it to launchpad. Thanks!

didrocks:
Perfect! tell me what you think about my answers. I'm going to work soon on the dartlang support, but I think we can cut a release with your changes + the appmenu by default to promote it!

Nice work again ;)

Tinche:
Hi,

* medium tests: OK, I'll look into it. We already have kinda complex handling of query params in the test server, and I'll need to add to this (since the only difference between Idea Community and Idea Ultimate is in the query params). I think since '?' and '&' are valid file path characters, just short-circuit the logic if the exact file name is found, and add files like "../download_thanks.jsp?edition=IU&os=linux"

* commands: the problem is staying consistent. Technically only Idea has an 'ultimate' version (I guess because you can install *all* the other plugins into it), PyCharm's commercial edition is called 'Professional'. Maybe go with idea-ultimate, pycharm-professional, rubymine-professional, webstorm-professional, etc?

* desktop files: I guess I misspoke when I put this into 'open points', I think it's fine with these changes. I will add the missing startupClass even though for me it works fine without. Heh, I didn't actually know the Unity Dash matches on comments. Is there any documentation on making good desktop files? Anyway, this item is not a problem for me.

umake/frameworks/ide.py
Tinche: It's to have the exact same name as the tools themselves generate. For example, Idea Ultimate has this wizard that runs when it's started up for the first time, and it can overwrite our .desktop file, so this way the name won't change.

umake/frameworks/ide.py
Tinche: Again, to match their file.

didrocks:
Hey!

* medium: sure, please go ahead with this.

* commands: I like your idea to really follow the product name, so "idea-ultimate, pycharm-professional, rubymine-professional, webstorm-professional", this makes sense to me

* desktop files: unfortunately, no good documentation that ensures being well integrated between GNOME, Unity, KDE, XFCE… You can just run "desktop-file-validate" on them to ensure you are not violating any rules :)

Tinche:
Working on this a little more, having problems not having jayatana in the Docker container.

So I thought I'd add it to create_packages.sh, but this doesn't actually work since the package isn't available so apt-cache can't extract the version. Now I could probably hack around this, but a different question: how is this supposed to work in practice, on pre-vivid?

If jayatana is missing, Android Studio/Jetbrains don't get offered at all, and as far as I can tell Ubuntu Make can't add the PPA itself. Should we implement adding the PPA on pre-vivid?

didrocks:
Good catch! I'm still waiting for the desktop vm to be up again by IS to get the medium tests running and being able to catch that.

jayatana is the first package that's not part of the archive in pre-vivid and part of the fake repository. The version isn't as important (and we can force it if it is), so I just fill a version if we didn't find any on the repository (I don't think it worths adding the ppa for this). The fix is in https://github.com/ubuntu/ubuntu-make/commit/016153bdcf4047160ddcb826b7ec9b6dcc195a54, and then, I added jayatana to fix the medium tests that are now all passing in master: https://github.com/ubuntu/ubuntu-make/commit/8c8ba3847b917e88a4f357aa3d54f57bb9227897. I also pushed to dockerhub the new image that you can use!

Sorry for this oversight ;)

Tinche:
Should be good to go.

didrocks:
Excellent! I'll make the change that we just discussed, otherwise, everything works well, including medium tests.

A big thank as usual :)

Tinche:
Sweet, thanks for reviewing!

PR #66, opened 2015-01-21
Update ide.py

Tinche:
I think Eclipse doesn't actually need jayatana, right?


didrocks:
Oh, you are indeed right. This was previously (some cycles ago) only available in the ubuntu version as a distro patch, but seems it's upstream (just tested luna and Dart Editor), maybe the assets need to be changed for the medium test as well? (and so, we need specific eclipse and intellij assets), mind looking at this?

Nice pick! :)

Tinche:
Fixed elsewhere.

Issue #67, opened 2015-01-21
Intellij IDEA: Can't parse the download URL from the download page.


ChristopherBasinger:
##### Need to know how to find the script that contains either the <code>wget</code> or <code>curl</code> command for Intellij idea.
```shell
sudo apt-get purge ubuntu-developer-tools-center
sudo apt-get clean && sudo apt-get autoremove
sudo apt-add-repository ppa:ubuntu-desktop/ubuntu-make
sudo apt-get update && sudo apt-get install ubuntu-make
cd ~
devandroid@christopher-Lenovo:/usr/local/lib/python3.4/dist-packages$ umake ide idea
Choose installation path: /home/devandroid/tools/ide/idea
ERROR: Can't parse the download URL from the download page.

```
I was unable to find where the python script that was specified by an include statement at the top of the shell script for running the ubuntu-make application by executing its' 
```shell
umake
```
```shell
/usr/bin/umake
```

alias which mapped to 

```shell
cat /usr/bin/umake 
```

The umake script had the following contents:
```python
  GNU nano 2.2.6                                  File: /usr/bin/umake                                                                          
#!/usr/bin/python3
# EASY-INSTALL-ENTRY-SCRIPT: 'Ubuntu-Make==0.0.1','console_scripts','umake'
__requires__ = 'Ubuntu-Make==0.0.1'
import sys
from pkg_resources import load_entry_point
if __name__ == '__main__':
    sys.exit(
        load_entry_point('Ubuntu-Make==0.0.1', 'console_scripts', 'umake')()
    )
```
The JetBrains page for downloading the Community Edition idea package looks to be different than before. Previously I don't recall the option to choose my protocol provided via two separate URIs for both ```shell
```shell
http://download.jetbrains.com/idea/ideaIC-14.0.2.tar.gz
https://download.jetbrains.com/idea/ideaIC-14.0.2.tar.gz
```
protocol and/or at the least specifying that there was an option. Previously I had tried in the past to download Jet Brains package using 
```shell
$ wget http://download.jetbrains.com/idea/ideaIC-14.0.1.tar.gz
```
and whenever I'd added the two parameters following the command
```shell
wget --no-cookies --no-check-certificate https://download.jetbrains.com/idea/ideaIC-14.0.1.tar.gz
```
I would end up having to remove the **s** from the protocol due to having inserted it myself. The 
```shell
wget --no-check-certificate http://download.jetbrains.com/idea/ideaIC-14.0.1.tar.gz
[...]

PR #68, opened 2015-01-21
Idea fix

Tinche:
Fix for #67.

Also removed jayatana from Eclipse (also for Docker) and back-ported the new local-server.py feature from the Idea Ultimate branch.


Tinche:
After this is merged I'll rebase the Idea Ultimate branch on it and that should be ready too.

didrocks:
Looking good, tested and works well. Thanks for the prompt reaction! :)

I hope the HTTPS tag will be more stable in their download page. I'm doing a release with this fix as I'm writing and blog about it. Let's plan the idea ultimate fix on this later on.

PR #72, opened 2015-01-30
FTP adapter file header.

Tinche:
Forgot to add this.


umake/network/ftp_adapter.py
```diff
@@ -1,3 +1,22 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2014 Canonical
```
didrocks: We are in 2015 I guess ;) No biggie anyway!

didrocks:
Perfect, thanks a lot, easy review ;)

Issue #98, opened 2015-05-04
Man pages for umake


johnsamuelwrites:
It will be quite useful if there are man pages for umake. Current documentation of umake is only obtained by the following command
$ umake --help

$ man umake 
No manual entry for umake
See 'man 7 undocumented' for help when manual pages are not available.

didrocks:
Thanks for the feature request! Nice one, indeed, and can be easily dealt by anyone who wants to make a first contribution :)

PR #99, opened 2015-05-04
POD and Script for generating man pages from POD

johnsamuelwrites:
Man pages for umake (https://github.com/ubuntu/ubuntu-make/issues/98).


didrocks:
Thanks for this patch proposed shortly after the bug report :)

I'm a little bit vary about listing manually all supported frameworks, it's a good candidate to remain out of date very quickly, isn't it? :)

I think we should rather get that integrated into setup.py (when running the install command), to generate from bin/umake --help, changing some part (via regexp probably) to add proper tags as you did, and then, adding the synposis and license information.

Does it make sense? For any question or any additional support I can bring in Ubuntu Make to ease your task, do not hesitate to ask, I'm around :)

johnsamuelwrites:
That sounds a better idea. I think it's better to integrate into setup.py to have an exhaustive list (of frameworks) available during the install time

carlosperate:
Also, I don't see Visual studio code included

didrocks:
The man page is now generated from --help and this is in master! Do not hesitate if you think you can enhance this, but I'm closing this PR for now as any enhancement would need more work anyway :)

Issue #102, opened 2015-05-07
Arduino support


didrocks:
From https://bugs.launchpad.net/ubuntu/+source/ubuntu-make/+bug/1452440:

I think is a good idea if Ubuntu Make can install Arduino official IDE automagically. The version packaged is very old (1.0 in repository instead 1.6 from website) and is hard to put to work the website version (create a link for executable, create a .desktop file, etc.)

didrocks:
I guess this a good idea, maybe a simple thing for a new contributor?

Tinche:
I can take a look over the weekend if it's not hard... Been a while since
I've done anything :)

didrocks:
I would be delighted! I unfortunately don't have any arduino hardware to do a proper decision job on this :)

Tinche:
Ugh, me neither. I hope just downloading (without testing on real hardware)
will be enough to be useful for someone.

didrocks:
I think my manager has some hardware, I'll ask him to give it a try with your PR!

mhalano:
I don't think an Arduino board is really necessary for this task. The idea is just install the IDE, but there is a trick related with the serial communication, so the user need to be added in the 'dialout' group. But already exist a package .deb, so a look into that solve most problems.

didrocks:
This was is now released as of Ubuntu Make 0.9

PR #103, opened 2015-05-10
Arduino

Tinche:
Fix for #102.

Not yet ready for merging - need to write tests. It's feature complete though.

Had to implement some more functionality into the DownloadCenter - support for final_url (the final URL after all redirections), and cookie support for both requests and responses.

### DIFF
--- a/tests/large/test_ide.py
+++ b/tests/large/test_ide.py
@@ -407,3 +407,47 @@ class PhpStormIDETests(LargeFrameworkTests):
         self.expect_and_no_warn("PhpStorm is already installed.*\[.*\] ")
         self.child.sendline()
         self.wait_and_no_warn()
+
+
+class ArduinoIDETests(LargeFrameworkTests):
+    """The Arduino Software distribution from the IDE collection."""
+
+    TIMEOUT_INSTALL_PROGRESS = 120
+    TIMEOUT_START = 60
+    TIMEOUT_STOP = 60
+
+    def setUp(self):
+        super().setUp()
+        self.installed_path = os.path.expanduser("~/tools/ide/arduino")
+        self.desktop_filename = "arduino.desktop"
+
+    @property
+    def arch_option(self):
+        """we return the expected arch call on command line"""
+        return platform.machine()
+
+    def test_default_install(self):
+        """Install the distribution from scratch test case"""
+        self.child = pexpect.spawnu(self.command('{} ide arduino'.format(UMAKE)))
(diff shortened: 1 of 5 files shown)


Tinche:
Also, going by @mhalano's comment and http://playground.arduino.cc/linux/ubuntu, looks like we need to depend on gcc-avr and avr-libc, and add the user to the dialout group.

mhalano:
@Tinche Dependencies and add to group is a problem which can be solved?

Tinche:
Of course :) Just need to get around to it.

didrocks:
Looking great! Nice work Tinche ;)

Yeah, adding to the group is fine, maybe, we need to add a hook for requiring root:
- if some packages are required and not installed, root will be required (this is already handled).
- maybe we can add a hook from the BaseFramework: look at self.need_root_access in __init__(). I would call a generic (doing a no-op in BaseFramework) hook. Then, in the arduino framework, using that hook to requiring root if the user isn't in the dialout group. I'm happy to help on this if needed.

Adding some more comments inline for this first review before testing.

Tinche:
Thanks for the comments, @didrocks ! I still need to get around to the group adding hook and tests. Might take a little while, it's the middle of a busy work week :)

didrocks:
No hurry at all (I won't be available on Thursday anyway) ;) Great work!

Tinche:
So, the status of this issue is what everything is done except:

* adding the user to the dial group (maybe you could take a look, @didrocks, since you had an idea of how to do this in a reusable manner)
* the medium (Docker) test

I've put in some work on the medium test, but I can't get it to pass since the large test it's based on depends on accessing the desktop file inside the container (_get_path_from_desktop_file, assert_exec_exists, assert_icon_exists). I see other medium tests use these, so I'm guessing the other medium tests don't work either?

Honestly, when I'm working on adding something new, I spend a disproportionate amount of time on the medium test alone. At this point I'm wondering if it's worth it.

didrocks:
Looking good! (just look at your recent commits)

I guess they still worths it, unfortunately, nothing in Ubuntu CI is currently capable of running them. I'll still work on those, don't worry :)

I'm on a sprint next week, so I may be able to start looking at this, but I guess the final touch will be the following wekk. I'll fork your branch and add the group thingy.

Thanks again for your hard work!

didrocks:
I've fixed all medium test issues in trunk, merging now your PR and adding the medium test + dialout group addition!

Thanks again for your work, I'll probably release it after my holidays, so staging it a little bit in trunk first!

Issue #80, opened 2015-04-05
No option to install Firefox Developer Edition in other languages than English


marcinant:
As in topic. 

Firefox Developer Edition is available in localized versions. It could be nice to have an ability to install localized version with umake.

didrocks:
Good idea! Seems like that can be an easy contribution, let's open that to any volounteer :)

didrocks:
This one is implemented as of Ubuntu Make 0.9.2 with a lang question and a --lang= for command-line only.

PR #119, opened 2015-06-29
Add support for non-english firefox-dev installations

oijazsh:
A possible solution for issue #80. I was wondering whether the I have made changes are reasonable.

I was also unable to perform the medium tests as I was not sure how to generate the required certificate.


didrocks:
First, sorry for the long delay for this answer, I was on holidays and when I got back, I wanted to clear the ground for releasing 0.9 before being able to review this PR with a clear mind (and no jetlag ;)).

Thanks a lot for your work! This is fantastic and a great added value for Ubuntu Make. Really nicely done as well thanks to the Choice() element and a default (US) which makes sense.

Some general enhancements that I think can be done:
1. maybe write what we use as the default for the language choice (or should we do that in a more generic way?) for the choose language string?

2. when I enter an invalid entry, the other message is:
[ach/af/sq/ar/an/hy-AM/as/ast/az/eu/be/bn-BD/bn-IN/brx/bs/br/bg/my/ca/zh-CN/zh-TW/hr/cs/da/nl/en-GB/en-ZA/en-US/eo/et/fi/fr/fy-NL/ff/gd/gl/de/el/gu-IN/he/hi-IN/hu/is/id/ga-IE/it/ja/kn/ks/kk/km/kok/ko/lv/lij/lt/dsb/mk/mai/ms/ml/mr/nb-NO/nn-NO/oc/or/fa/pl/pt-BR/pt-PT/pa-IN/ro/rm/ru/sat/sr/si/sk/sl/son/es-AR/es-CL/es-MX/es-ES/sv-SE/ta/te/th/tr/uk/hsb/uz/vi/cy/xh] cfdsfds
ERROR: Your entry 'cfdsfds' isn't an acceptable choice. choices are: [] and ['ach', 'af', 'sq', 'ar', 'an', 'hy-AM', 'as', 'ast', 'az', 'eu', 'be', 'bn-BD', 'bn-IN', 'brx', 'bs', 'br', 'bg', 'my', 'ca', 'zh-CN', 'zh-TW', 'hr', 'cs', 'da', 'nl', 'en-GB', 'en-ZA', 'en-US', 'eo', 'et', 'fi', 'fr', 'fy-NL', 'ff', 'gd', 'gl', 'de', 'el', 'gu-IN', 'he', 'hi-IN', 'hu', 'is', 'id', 'ga-IE', 'it', 'ja', 'kn', 'ks', 'kk', 'km', 'kok', 'ko', 'lv', 'lij', 'lt', 'dsb', 'mk', 'mai', 'ms', 'ml', 'mr', 'nb-NO', 'nn-NO', 'oc', 'or', 'fa', 'pl', 'pt-BR', 'pt-PT', 'pa-IN', 'ro', 'rm', 'ru', 'sat', 'sr', 'si', 'sk', 'sl', 'son', 'es-AR', 'es-CL', 'es-MX', 'es-ES', 'sv-SE', 'ta', 'te', 'th', 'tr', 'uk', 'hsb', 'uz', 'vi', 'cy', 'xh']

For android studio license agreement for instance, the message is:
[I Accept (a)/I don't accept (N)] fdfdsfdsfds
ERROR: Your entry 'fdfdsfdsfds' isn't an acceptable choice. choices are: ['a', 'N'] and ['I Accept', "I don't accept"]

I guess we can avoid the empty []. Note that the error is duplicated twice, it seems to be an issue with the generic framework, I'm happy to give both issues a look if you didn't get a chance at it.

3. We have some automated non interactive mode for auto accepting license on frameworks which requires more inputs. That enables people to use ubuntu make in scripts.
For instance: "$ umake android ~/android-studio --accept-license" will install android studio without any question in /home/<user>/android-studio.
Maybe we should do something similar like --lang=<lang_name>?
The accept-license option is automatically generated for any framework which declares needing a license. I don't think we should do something generic for this case but just implement it into the framework.

What do you think about those proposals?

On the testing side, I think that the large tests can grow with an additional one to select another language than the default and checking (maybe with some installed file being present?) that this language version has been indeed installed.
The current one can also check some files to ensure that en_US is selected.
A third one would be nice as well to test the --lang= option.

For the medium tests, I'm happy to generate the certificate for www.mozilla.org. If you are interested, the instructions to create one are available at https://github.com/ubuntu/ubuntu-make/blob/master/tests/tools/local_server.py#L36. I just wrote them to remember and it's been prooved to be quite useful ;) (I would just wipe the .crt and .pem for download.mozilla.org and add those). Remember then you need to modify the Dockerfile to reference the new .crt file and regenerate the docker image with "$ docker build .". Again, I'm happy to deal with this (I will have to rebuild and push the docker image anyway to the dockerhub) if you are busy enough. ;)
If you add another test for a local != en_US, remember as well to ship a mock content for that language (respecting the additional file you are checking to ensure the right locale).


Phew! That's why I wanted to wait a little bit before making this review :)
In a nutshell, excellent work, I'm quite impressed by how you understood the whole framework, nicely done! Do not hesitate to ask if you need any help for the remaining suggestions (and feel free to share your thoughts about them too). As you can see from the review itself, I just have a minor typo comment, otherwise everything looks and work well from my perspective! (tried en_US and fr)!

Keep me posted, and thanks again for this valuable contribution!

umake/frameworks/web.py
```diff
@@ -49,21 +50,54 @@ class FirefoxDev(umake.frameworks.baseinstaller.BaseInstaller):
...
+    @MainLoop.in_mainloop_thread
+    def language_select_callback(self, url):
+        url = url.replace("&amp;", "&")
+        logger.debug("Found download link for {}".format(url))
+        self.download_requests.append(DownloadItem(url, None))
+        self.start_download_and_install()
+
+    @MainLoop.in_mainloop_thread
+    def get_metadata_and_check_license(self, result):
+        """Diverge from the baseinstallator implementation in order to allow language selection"""
```
didrocks: Just a small typo: baseinstaller instead of baseinstallator

oijazsh:
Thanks for the review and for pointing out the instructions for generating the certificates (and for pointing out the typo)! I was able to successfully run the medium test (it passes).

With regards to your proposals, I would be happy to try my hand at them. I am guessing that simply writing the default into the choose language string should be reasonable enough. The error message duplication seems simple enough to fix – I should be able to do it today.

For expanding the tests, it appears that the Firefox archives contain files by the name of {language}[.aff, .dic] in the  dictionaries folder. I suppose I could simply for check for the existence of one of those to verify the installed version? If so I believe I could handle adding the additional tests and the –lang=? option soon enough.

didrocks:
Nice that the medium tests are now passing (they are not always the most straightforward to run) and that generating the certificates worked!

Your couple of commits looks good. Writing the default into the choose language string, as you did is good enough IMHO :)
I saw as well that you found the issue with the empty [] due to having shortcuts. It seems that existing tests should cover that already. You can enable test coverage in runtests and run tests/small/test_interactions.py, TestInteractions class to ensure it, but I check there that we just raise an exception.
Thanks as well for fixing the double erroring!

Just a question, why did you change the ID for Choices from the string to be an enumeration index? (both are fine to me, I'm just curious ;))

On expanding tests: agreed with that strategy. Think about updating the mock assets as well to match! 

Thanks again for your good work, keep me posted :)

oijazsh:
My only reason for setting it to an enumeration index was to be more consistent with other places where the TextWithChoices interaction was used (namely, the tests, LicenseAgreement and YesNo); I had no particularly strong reason for doing it either way.

I've added in the --lang=? functionality and set up the large tests and mock contents for the medium tests. Both medium and large tests currently pass.

According to the coverage test, the code in TextWithChoices.choose() is still covered. However, 4 of the tests were now failing since we now do not need to expect an error; I have changed those tests accordingly.

didrocks:
Excellent work! Just again a very minor nitpick corresponding to logging.
The only drawback I see is that you didn't choose french for testing! (kidding ;))

I've run the large and medium tests (after rebuilding the docker image), looked at the medium assets, and everything looks good! pep8 is passing as well. I'm quite thrilled that you got as well how to add the option for that particular framework :)

Great small tests fixes, but the change in previous commit removing the exception introduced a minor regression. If you:
"./bin/umake web firefox-dev /tmp/somewhere --lang=pppp"
You get:
ERROR: Could not find a download url for language pppp

However, as we don't raise an exception, umake exits with 0 and not in error, where I guess it should. Maybe we should add a test to ensure it exits 1, reintroduce raising the exception and rather remove the error output? (as raising the exception would print the error). I can give it a look to that if you don't have time.

Keep me posted, we are close to merge your changes!

oijazsh:
That is indeed leftover from debugging – real sorry about that! I'll replace it in my next commit.

I am not sure I understand what you mean by the regression however. I understand that the expected functionality is for an error code of 1 to be returned when exiting due to an error. However, as far as I can tell, the mistake I made is calling UI.return_main_screen() without any arguments instead of UI.return_main_screen(1). I also did not remove any exception raising statement I believe. I <b>did</b> remove a logger.error() call in TextWithChoices but that interaction is now not used when --lang is specified. Am I overlooking something?

didrocks:
Sorry, it's been a long time I wrote this apart and I was puzzled, indeed, that error is managed and so logged as error: it won't result in automatically exiting the process. It's for unhandled exception where we log as well the error and exits the process with 1 as an error exit code. :)

So everything is good apart as you told that you need to UI.return_main_screen(status_code=1) and add a simple test for this? Then, good for merging! (I'll cut a release next week with your feature!). Thanks again!

oijazsh:
Alright – thanks for helping me through this pr! I just have a few questions that came to my mind as wrote the rather simple test but since those are not directly related to pull request I will ask them in a separate issue.

didrocks:
Excellent! Thanks again for your work :)

I'll run tests a last time on Monday, merge and make a release! I know I repeat myself, but you have done an excellent job on this, rock on! :)

didrocks:
Everything is fine, let's merge it now! I'll let an automated test runner pass and the release should be ready for tomorrow. Nice work again, big thanks!

Issue #120, opened 2015-06-29
Add Eclipse Mars support


syndbg:
Latest supported by ubuntu-make is Luna.

https://projects.eclipse.org/releases/mars

marcinant:
Does it support upgrade of existing instances?

didrocks:
@marcinant: it does when upstream provides their own update mechanism (contrary to the ubuntu package, we are compatible with the way upstream autoupdate their tools). So, if eclipse does provide so, it will work.

For tools that don't offer upgrade mechanism, there is a discussion on #74.

didrocks:
This is now available as of Ubuntu Make 16.01. Latest eclipse release should now always be delivered with this refactoring!

PR #121, opened 2015-06-29
[#120] Added Eclipse Mars IDEs support and refactoring

syndbg:
Addresses #120 

Currently my Docker VM is terribly slow. So it'll take one more day until I have all the tests to compliment the functionality.

Last thing left to do: add server-content data. This depends on me setting up the Docker VM and generating md5 sums


Tinche:
I think probably the default Eclipse installation should be the newest one. If someone wants Luna explicitly, we can offer 'eclipse-luna', but just 'eclipse' should install Mars now.

syndbg:
Just looked again at https://eclipse.org/downloads/packages/release/Mars/R and https://eclipse.org/downloads/packages/release/Luna/R 

I'm not all too familiar with Eclipse's release policies, but Mars has no Standard Edition so far. That's a bit troublesome.


At least I can add Eclipse Java EE support.

syndbg:
Ok I haven't been active here in the last few days. 
My status is that I'm adding support for JEE, Java, C/C++ and PHP editions support.
It'll take some more time since I'm not focused into this.

didrocks:
Sorry Anthony, I have been on holidays until today, hence my lack of anwser :)

I agree with Tin, the Eclipse framework should always try to install latest available version of it. So, updating it to Mars for now seems to make sense? (If someone still want Luna, we can add it as an additional framework, but I'm unsure I feel confortable with this, as it's breaking the "we always download latest and greatest" via Ubuntu Make).

Let's wait as you told for them to grow Mars into the standard edition. And indeed, feel free to add support for the other non Java-related Eclipse eidition!
Thanks for handling this, do not hesitate if you have any question or need a hand.

syndbg:
Progress update. 

I added JEE, Java, C/C++ and PHP editions support.
However during download I get #123 . The weird thing is that it works fine for the (pre-PR) Eclipse Luna SE download.

Tonight I'll push my local work to the branch so you can see my code exactly as it is.

didrocks:
Good news! Maybe its really time to fix #123 with capping it to 100% (would be interesting to get the --verbose option to see what's the size of expected and current download to see why we exceed it beforehand though).

Ping me as soon as you push your changes, thanks again! :)

syndbg:
After the progress bar fix, new issue:

```

ERROR: A decompression to /root/tools/ide/eclipse-ee failed: list index out of range                                  |
  File "/usr/lib/python3.4/concurrent/futures/thread.py", line 54, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/dev-umake/umake/decompressor.py", line 111, in _decompress
    dir_path = glob(os.path.join(dest, dir))[0]
IndexError: list index out of range
ERROR: list index out of range
```

runtests
```diff
@@ -19,7 +19,10 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 
 import argparse
-import configparser
+try:
+    import configparser
+except ImportError:
+    import ConfigParser as configparser
```
syndbg: Needed to handle both cases of using the old or new ConfigParser

runtests
```diff
@@ -136,7 +139,7 @@ if __name__ == '__main__':
         import tests
         tests.tools.set_local_umake()
 
-    if not "all" in args.tests and len(args.tests) > 0:
+    if "all" not in args.tests and len(args.tests) > 0:
```
syndbg: Just a flake8 complaint

tests/large/test_ide.py
```diff
@@ -31,26 +31,25 @@
 logger = logging.getLogger(__name__)
 
 
-class EclipseIDETests(LargeFrameworkTests):
-    """The Eclipse distribution from the IDE collection."""
-
+class BaseLargeIdeTest(LargeFrameworkTests):
     TIMEOUT_INSTALL_PROGRESS = 120
```
syndbg: Used to reduce the duplication in Large IDE tests

tests/large/test_ide.py
```diff
@@ -31,26 +31,25 @@
...
     TIMEOUT_STOP = 60
 
     def setUp(self):
         super().setUp()
-        self.installed_path = os.path.expanduser("~/tools/ide/eclipse")
-        self.desktop_filename = "eclipse.desktop"
+        self.name = ""
+        self.command_name = ""
+        self.installed_path = ""
+        self.desktop_filename = ""
```
syndbg: I personally dislike the need for `self.command_name`. I couldn't dig the source code where the Installer's name, ex. Eclipse EE, gets converted to eclipse-ee. 

Also the fact that we have to manually give strings that at some point will differ than the actual implementation and give   misleading truthy asserts, is a good reason for me to ask - why aren't we instantiating installers such as Eclipse and directly using their properties? The later is the optimal TDD approach but I think there might be a performance reason why we are not doing it?

umake/frameworks/baseinstaller.py
```diff
@@ -199,18 +203,23 @@ def get_metadata_and_check_license(self, result):
...
-            if license_txt.getvalue() != "":
-                logger.debug("Check license agreement.")
-                UI.display(LicenseAgreement(strip_tags(license_txt.getvalue()).strip(),
-                                            self.start_download_and_install,
-                                            UI.return_main_screen))
-            elif self.expect_license and not self.auto_accept_license:
-                logger.error("We were expecting to find a license on the download page, we didn't.")
-                UI.return_main_screen()
-            else:
-                self.start_download_and_install()
```
syndbg: A flake8 complaint about method complexity. I would've wanted to reduce it further but I couldn't.

I suppose this will do for now.

umake/frameworks/baseinstaller.py
```diff
@@ -237,34 +246,34 @@ def get_progress(self, progress_download, progress_requirement):
...
         if not self.pbar.finished:  # drawing is delayed, so ensure we are not done first
             progress = self._calculate_progress()
             self.pbar.update(progress)
 
+    def _compute_balanced_requirement(self):
+        if not self.pkg_to_install:
+            self.balance_requirement_download = 0
+            self.last_progress_requirement = 0
+            if self.last_progress_download is None:
+                return
```
syndbg: `_compute_balanced_requirement`'s two return statements are kind of stupid way to refactor this.

I'm pretty certain I could just throw a `ReturnMainLoop` exception to stop the loop without doing the even more redundant

```python
if self._compute_balanced_requirement() is None:
    return
```

umake/frameworks/baseinstaller.py
```diff
@@ -237,34 +246,34 @@ def get_progress(self, progress_download, progress_requirement):
...
+                                                     self.total_download_size),
+                                                    0.15)
+
     def _calculate_progress(self):
         progress = self.balance_requirement_download * self.last_progress_requirement +\
             (1 - self.balance_requirement_download) * self.last_progress_download
         normalized_progress = max(0, progress)
-        normalized_progress = min(normalized_progress, 100)
-        return normalized_progress
+        return min(normalized_progress, 100)
```
syndbg: There was no need to reassign it again. Just a itsy-bitsy (personal) code style fix.

umake/frameworks/ide.py
```diff
@@ -53,7 +53,7 @@ def _add_to_group(user, group):
             output = subprocess.check_output(["adduser", user, group])
             logger.debug("Added {} to {}: {}".format(user, group, output))
             return True
-        except subprocess.CalledProcessError as e:
+        except subprocess.CalledProcessError:
```
syndbg: Not used, although I think the error might be useful?

umake/frameworks/ide.py
```diff
@@ -110,24 +106,23 @@ def done(download_result):
 
             logger.debug("Preparing to download the main archive.")
             if arch == 'i686':
-                download_url = self.DOWNLOAD_URL_PAT.format(arch='', suf='')
+                download_url = self.download_page_url.format(arch='', suf='')
             elif arch == 'x86_64':
-                download_url = self.DOWNLOAD_URL_PAT.format(arch='-x86_64',
-                                                            suf='')
+                download_url = self.download_page_url.format(arch='-x86_64',
```
syndbg: Kept the naming consistent as the `download_page_url` class variable in `BaseJetBrains`

umake/frameworks/ide.py
```diff
@@ -138,12 +133,264 @@ def is_installed(self):
         # check path and requirements
         if not super().is_installed:
             return False
-        if not isfile(join(self.install_path, "eclipse")):
+        if not isfile(join(self.install_path, "bin", self.executable)):
```
syndbg: Just uncoupling stuff so it works for all Eclipse IDEs

syndbg:
The latest commit is ready for review.

My only issue is with the Eclipse Automotive IDE, which I cannot download due to a checksum error. I think it's a falsy reaction. Actually it might be due to the following error when I try to download it from the browser. 

```
403 Forbidden

Sorry, this type of request is for Eclipse committers only. Please log in if you are a committer and wish to download this file.
```

If Eclipse Automotive is available for download only when logged in as a committer, should we still add support for it?

runtests
didrocks: I don't think this commit is necessary. We only support Ubuntu Make with python3, not pythoN2, so you can remove this try/except statement.

runtests
didrocks: Can you move all flake8/enhancement not related to this pull request in another PR? I prefer to have separated branches (I guess you can do a rebase and split that in 2)

tests/large/test_ide.py
didrocks: ack

tests/large/test_ide.py
didrocks: To answer your comment: even if it can be seen as not as optimal to duplicate some infos here, keep in mind that those tests are large tests. They are independant of any information that umake is providing (they just set in stone some user experience). So, we want to avoid loading anything computed by the umake modules and really fake user's typing, without using any of their properties.

Also note that the runner can be on another machine that the destination, and this one could have some frameworks that are not loaded in memory as requirements are not available.

tests/large/test_ide.py
```diff
@@ -31,26 +31,25 @@
...
     @property
     def arch_option(self):
         """we return the expected arch call on command line"""
         return platform.machine()
 
-    def test_default_eclipse_ide_install(self):
-        """Install eclipse from scratch test case"""
-        self.child = pexpect.spawnu(self.command('{} ide eclipse'.format(UMAKE)))
+    def test_default_install(self):
+        self.child = pexpect.spawnu(self.command('{0} ide {1}'.format(UMAKE, self.name)))
```
didrocks: Please keep the coherence with the rest of the umake style and don't create any uncessary diff: when we only have few arguments to a string, only use {}, no {index}.

tests/large/test_ide.py
```diff
@@ -65,37 +64,19 @@ def test_default_eclipse_ide_install(self):
...
                                 stderr=subprocess.DEVNULL)
 
-        # on 64 bits, there is a java subprocess, we kill that one with SIGKILL (eclipse isn't reliable on SIGTERM)
-        if self.arch_option == "x86_64":
-            self.check_and_kill_process(["java", self.arch_option, self.installed_path],
-                                        wait_before=self.TIMEOUT_START, send_sigkill=True)
-        else:
-            self.check_and_kill_process([self.exec_path],
-                                        wait_before=self.TIMEOUT_START, send_sigkill=True)
+        self.check_and_kill_process(["java", self.installed_path], wait_before=self.TIMEOUT_START)
```
didrocks: Mind expanding a bit? Why is the 32/64 bits isn't different anymore? Did you try it on 32 bits?

tests/large/test_ide.py
```diff
@@ -65,37 +64,19 @@ def test_default_eclipse_ide_install(self):
...
         proc.wait(self.TIMEOUT_STOP)
 
         # ensure that it's detected as installed:
-        self.child = pexpect.spawnu(self.command('{} ide eclipse'.format(UMAKE)))
-        self.expect_and_no_warn("Eclipse is already installed.*\[.*\] ")
+        self.child = pexpect.spawnu(self.command('{0} ide {1}'.format(UMAKE, self.command_name)))
+        self.expect_and_no_warn("{0} is already installed.*\[.*\] ".format(self.name))
```
didrocks: Same, about {} and diff, please revert this, apart from changing the command name.

While you are doing that, as they are all on the same category, maybe add a category_name = ide and change it?

tests/medium/test_ide.py
```diff
@@ -50,12 +50,203 @@ class EclipseIDEInContainerFTP(ContainerTests, test_ide.EclipseIDETests):
     def setUp(self):
         self.hostname = "www.eclipse.org"
         self.port = "443"
-        self.ftp = True
         # we reuse the android-studio repo
-        self.apt_repo_override_path = os.path.join(self.APT_FAKE_REPO_PATH, 'eclipse')
+        self.apt_repo_override_path = os.path.join(self.APT_FAKE_REPO_PATH, 'android')
```
didrocks: Nicely spotted!

umake/decompressor.py
```diff
@@ -28,7 +28,6 @@
 import tarfile
 import zipfile
 
-
```
didrocks: If you want to do this kind of cleanage, please do it in one commit, and in another branch. That doesn't have anything to do with the current matter.

umake/frameworks/baseinstaller.py
```diff
@@ -19,7 +19,6 @@
 
 
 """Downloader abstract module"""
-
```
didrocks: ditto

umake/frameworks/baseinstaller.py
```diff
@@ -125,6 +124,7 @@ def confirm_path(self, path_dir=""):
             return
 
         logger.debug("Installation path provided. Checking if exists.")
+
```
didrocks: I would prefer to keep the extra line here for readablity

umake/frameworks/baseinstaller.py
```diff
@@ -137,6 +137,7 @@ def confirm_path(self, path_dir=""):
                     UI.display(YesNo("{} isn't an empty directory, do you want to remove its content and install "
                                      "there?".format(path_dir), self.set_installdir_to_clean, UI.return_main_screen))
                     return
+
```
didrocks: I would prefer to keep the extra line here for readablity

umake/frameworks/baseinstaller.py
```diff
@@ -173,21 +174,24 @@ def get_metadata_and_check_license(self, result):
             logger.error("An error occurred while downloading {}: {}".format(self.download_page, error_msg))
             UI.return_main_screen()
 
-        url, checksum = (None, None)
-        with StringIO() as license_txt:
+        with StringIO() as license_text:
+            url, checksum = (None, None)
```
didrocks: Same with this, the change itself is fine, but has nothing to do with this PR. Please get that in logical commits and in a separate PR.

umake/frameworks/baseinstaller.py
```diff
@@ -173,21 +174,24 @@ def get_metadata_and_check_license(self, result):
...
             in_license = False
             in_download = False
+
```
didrocks: For this (and following space lines): they are there for readibility, please do not stash them. I'll comment if there are occurences of some that I think make sense to stash

umake/frameworks/baseinstaller.py
```diff
@@ -173,21 +174,24 @@ def get_metadata_and_check_license(self, result):
...
             for line in result[self.download_page].buffer:
                 line_content = line.decode()
 
                 if self.expect_license and not self.auto_accept_license:
-                    in_license = self.parse_license(line_content, license_txt, in_license)
+                    in_license = self.parse_license(line_content, license_text, in_license)
```
didrocks: The rest of the framework is using txt for text, please do not change it

umake/frameworks/baseinstaller.py
didrocks: That's good for now IMHO (but as usual, in another PR and separate commits, please)!

umake/frameworks/baseinstaller.py
didrocks: That's not the case. Remember that downloads are threaded, we have:
- one thread (which is actually forked in a subprocess) to download the apt installation
- one thread per download items in progress.

Depending on which one answer first, we may still not have any pkg_to_install when the download items get its first progress to download, and so, we need to hang on at 0 until we get this information (we don't want to see the progress handler to jump backward). We are then in:
if self.last_progress_download is None or self.last_progress_requirement is None:
    return
-> The whole function returns None in that case.

Maybe add a comment about it in the _compute_balanced_requirement() doc?

umake/frameworks/baseinstaller.py
didrocks: I always dislike when there is
foo = changefrom(bar)
return anotherchangefrom(foo)

I think this removes heavily readability for very small perf gain, and I prefer your first version.

umake/frameworks/ide.py
didrocks: Yeah, add it to logger.error please (but same, in the other "refactoring" PR and in a separate commit).

umake/frameworks/ide.py
didrocks: That's a good one. Here for the commit separation, I would think:
- one commit introducing BaseEclipse, adding the new property + the rename
and that's it! :)

This kind of change can be in the "add Eclipse Mars" PR, as it's directly related to it.

umake/frameworks/ide.py
```diff
@@ -110,24 +106,23 @@ def done(download_result):
...
         DownloadCenter(urls=[DownloadItem(md5_url, None)], on_done=done, download=False)
 
     def post_install(self):
-        """Create the Luna launcher"""
         icon_filename = "icon.xpm"
         icon_path = join(self.install_path, icon_filename)
-        exec_path = '"{}" %f'.format(join(self.install_path, "eclipse"))
-        comment = _("The Eclipse Luna Integrated Development Environment")
+        exec_path = '"{}" %f'.format(join(self.install_path, self.executable))
+        comment = _("The {0} Integrated Development Environment").format(self.description)
```
didrocks: {0} -> {} (and same for followings)

umake/frameworks/ide.py
```diff
@@ -138,12 +133,264 @@ def is_installed(self):
...
+    """The Eclipse Foundation Luna distribution."""
+    download_page_url = "https://www.eclipse.org/downloads/download.php?" \
+                        "file=/technology/epp/downloads/release/luna/R/" \
+                        "eclipse-standard-luna-R-linux-gtk{arch}.tar.gz{suf}" \
+                        "&r=1"
+    executable = 'eclipse.sh'
+
+    def __init__(self, category):
+        super().__init__(name="Eclipse",
+                         description=_("Pure Eclipse Luna (4.4)"),
```
didrocks: Do not add any notion of Luna (that would be removed anyway? See my incoming general comment) or version in the .desktop file.

-> Most of tools autoupdate (I know, Eclipse doesn't really, but I hope this will come…) and we don't change the .desktop files automatically. So we prefer to keep the description functional as ("Eclipse for Java Developers"). Same comments for all other frameworks below.

umake/frameworks/ide.py
```diff
@@ -138,12 +133,264 @@ def is_installed(self):
...
+                         desktop_filename='eclipse.desktop',
+                         packages_requirements=['openjdk-7-jdk'])
+
+
+class EclipseJava(BaseEclipse):
+    """The Eclipse Foundation Mars for Java Developers distribution."""
+    download_page_url = "https://www.eclipse.org/downloads/download.php?" \
+                        "file=/technology/epp/downloads/release/mars/R/" \
+                        "eclipse-java-mars-R-linux-gtk{arch}.tar.gz{suf}" \
+                        "&r=1"
```
didrocks: As we are going to only support eclise mars I guess, please factorize the whole url in BaseEclipse class, and then, we only change the URL for flavor. Making sense?

umake/frameworks/ide.py
```diff
@@ -138,12 +133,264 @@ def is_installed(self):
...
+    executable = 'eclipse.sh'
```
didrocks: Same for executable -> seems mainly the same for all "eclipses" -> BaseEclipse class.

umake/network/download_center.py
```diff
@@ -34,6 +35,7 @@
 from umake.tools import ChecksumType
 
 logger = logging.getLogger(__name__)
+MAX_WORKERS = cpu_count()
```
didrocks: I don't really like executing code

Also, same comment than in most of part, please do not make any changes unrelated to the current PR, and separate those as well in different commits. (so good for another PR, but that one should be independant as well from the refactoring which should be a second one).

Now, on the change:
I don't like executing code when importing a module (apart from setting up loggers and small things like that). Mind moving that to tools.py? and reference it here?

umake/network/download_center.py
```diff
@@ -96,6 +98,7 @@ def __init__(self, urls, on_done, download=True, report=lambda x: None):
             else:
                 dest = BytesIO()
                 logger.info("Start downloading {} in memory".format(url_request))
+
```
didrocks: Please keep this blank line :)

umake/network/download_center.py
```diff
@@ -103,7 +106,7 @@ def __init__(self, urls, on_done, download=True, report=lambda x: None):
             future.add_done_callback(self._one_done)
 
     def _fetch(self, download_item, dest):
-        """Get an url content and close the connexion.
+        """Get an url content and close the connection.
```
didrocks: Argh, french/english :p Thanks for the typo fix! (should be in the refactoring/small fixes PR)

didrocks:
Hey, great work on your eclipse work! Seems like you spent quite some time on this :)

Here are some more general comments that I would like you to apply before resubmiting, I hope I didn't miss anything because the interleaved commits makes it hard to read sometimes.
- 2 commits for this work isn't enough, to be able to read easily the history, please separate them (using a git rebase) in more logicial chunks. Here, everything is really mixed and hard to follow.
- for instance I guess the test framework change (Adding a BaseLargeIdeTest is a good idea, but that should be a commit and a pull request on his own). Please separate it and submit it that way. Then, you can stash other commits on top of it in another pull request.
- please do not add support for python2, we won't support it (see ConfigParser part)
- do not change any general style that you can see in other frameworks, especially when you introduce disparities with the rest of the code. So no {0}, {1} when they are few parameters. Also, if you are doing such changes, please see my comment to put that in a PR and commits independent of the current one. So logical commits in small chunks (one for cleaning up all lines or typos, another one for some small refactoring) + merge request that makes sense == happy reviewer ;)
Please rework your proposal to follow those guidelines.

Do you mind to remind me why we still keep Eclipse Luna as this add Eclipse Mars? One of the core value of Ubuntu Make is to always deliver latest release to our users (I wasn't found of and get removed before merging mentions of "Luna" as we just want to get latest eclipse installed. We don't have the manpower to support latest versions + olders.

Finally, on your question on Eclipse Committers and co, I think we shouldn't add any auto installer support for things that requires you to be logged in on a website, so yeah, you can remove that one.
I am also unsure about the added value of those ones, do you mind expand (and if we keep them, I guess the description should be more appropriate for the user)?
- EclipseDsl
- EclipseRcpAndRap
- EclipseModelingTools
- EclipseAutomotive
- EclipseTesters
- EclipseParallel
- EclipseScout
-> Some may be interesting (modeling), but I'm afraid that we just lose a lot of users by adding unecessary complexily (imagine if we do the same with all netbeans ones!)

So, rather, let's select a few of them, that cover a different spectrum of developers (like java, C/C++, php), stick with those. Then, we only add more if we see demand arise, what do you think?

I think I gave your the great lines of my thoughts. To sum up, 4 pull requests should be in order (with logical commits please!):
- one to refactor the IDE large tests frameworks with a base class
- one for the number of thread we allow for parallel downloads
- another one for the small refactoring/cleanup
- and finally the eclipse-related ones.

Those chunks would take way less time to review and help you as well organizing your work (you can have that locally in one branch and git rebase heavily, but pushing that to separate PR and branches on github). The logical commits will help to read the changes/review here.

Thanks for your hard work here! Do not hesitate if you have any remark/questions :)

syndbg:
To answer some of your questions/comments:

> 2 commits for this work isn't enough, to be able to read easily the history, please separate them (using a git rebase) in more logicial chunks. Here, everything is really mixed and hard to follow.

I feel you. I don't know why I thought it's a good idea to put it all at once.

> ack, did you test the medium tests with that changes as I think it impacts them?

Couldn't. MacBook overheats and shutdowns at half the tests. I'm running them from a ubuntu docker container. Sad story is I have no other personal machine around.
 
> please do not add support for python2, we won't support it (see ConfigParser part)

Totally forgot that we don't support Python2. Bad Python hands, bad! 

> Do you mind to remind me why we still keep Eclipse Luna as this add Eclipse Mars? 

The Eclipse Luna is SE. There's no Eclipse Mars SE yet :fallen_leaf: . I hope there'll be a SE for Mars too although it a huge possibility it might not happen at all.

> Finally, on your question on Eclipse Committers and co, I think we shouldn't add any auto installer support for things that requires you to be logged in on a website, so yeah, you can remove that one.

Ack

> I am also unsure about the added value of those ones, do you mind expand (and if we keep them, I guess the description should be more appropriate for the user)?

I think it's good to have lots of choices.

> but I'm afraid that we just lose a lot of users by adding unecessary complexily (imagine if we do the same with all netbeans ones!)

I'm not sure I follow. Are there any hidden issues by adding every latest IDE edition? Netbeans has only 5 editions. Not that many, I think. :1234: 

> So, rather, let's select a few of them, that cover a different spectrum of developers (like java, C/C++, php), stick with those. Then, we only add more if we see demand arise, what do you think?

Sounds reasonable. I still wonder what the cons of adding them all are, but I can agree on this approach.
Luna - SE, Mars - Java, Java EE, C/C++, and PHP editions. Confirm?

> I think I gave your the great lines of my thoughts. To sum up, 4 pull requests should be in order (with logical commits please!):
> one to refactor the IDE large tests frameworks with a base class
> one for the number of thread we allow for parallel downloads
> another one for the small refactoring/cleanup
> and finally the eclipse-related ones.

Will do.

umake/network/download_center.py
syndbg: > Now, on the change:
> I don't like executing code when importing a module (apart from setting up loggers and small things like that). Mind moving that to tools.py? and reference it here

Will do. 

> Also, please note that it should be MAX_WORKERS -1 (as we download/install in parallel required packages from this same process)

Good catch!

umake/frameworks/ide.py
syndbg: Good idea. Will do.

didrocks:
Le 10/08/2015 21:29, Anton Antonov a écrit :
> To answer some of your questions/comments:
>

Thanks for your answers! My turn ;)

>
>     Do you mind to remind me why we still keep Eclipse Luna as this add
>     Eclipse Mars?
>
> The Eclipse Luna is SE. There's no Eclipse Mars SE yet :fallen_leaf: . I
> hope there'll be a SE for Mars too although it a huge possibility it
> might not happen at all.

Ok, then, let's put in the description of the "eclipse" framework just 
Eclipse SE (no mention of Luna).

For the Mars one, let's call it Eclipse EE. That sounds good?
>
>     I am also unsure about the added value of those ones, do you mind
>     expand (and if we keep them, I guess the description should be more
>     appropriate for the user)?
>
> I think it's good to have lots of choices.
>
>     but I'm afraid that we just lose a lot of users by adding unecessary
>     complexily (imagine if we do the same with all netbeans ones!)
>
> I'm not sure I follow. Are there any hidden issues by adding every
> latest IDE edition? Netbeans has only 5 editions. Not that many, I
> think. :1234:

By experience: too much choice without enough information to have an 
informed decision puzzle the user, and thus, make the whole benefit 
vanish. That's one of the reason Ubuntu itself was successful at launch 
compared to other distros: removing the whole first screen "which 
desktop environment do you want to use" and having sane defaults and few 
options at start (without blocking the possibility for more later).


>
>     So, rather, let's select a few of them, that cover a different
>     spectrum of developers (like java, C/C++, php), stick with those.
>     Then, we only add more if we see demand arise, what do you think?
>
> Sounds reasonable. I still wonder what the cons of adding them all are,
> but I can agree on this approach.
> Luna - SE, Mars - Java, Java EE, C/C++, and PHP editions. Confirm?

Yeah, let's call Luna - SE -> Eclipse - SE.
What's the difference between what is today Mars Java and what would be 
Mars SE? (I though the SE was the java based one)?

rockwalrus:
"Eclipse Standard" was renamed to "Eclipse for Eclipse Committers" in Mars, to better reflect what it's really for at this point.  So if you're waiting for Eclipse Standard to show up in Mars, you're going to be waiting a long time. :)

syndbg:
@rockwalrus waited a long time! :palm_tree: I'm going to finally address this in the following days.

didrocks:
Excellent! Keep us posted and do not hesitate if you need any help/have any question :) (for the testing side, maybe wait for @rockwalrus's PR to land first)

didrocks:
We now have some generic framework for eclipse merged in. However, some of your work here can be still really interesting if you are motivated on working on this (more refactorization). The 2 branches have diverged a lot, I suggest that you open another PR with those changes in, that would be easier! :)

Issue #123, opened 2015-07-08
Intermittent progressbar crash (ValueError: Value out of range)


pjdelport:
On a fresh installation of Ubuntu 15.04, after doing the usual system package updates, my first attempt at installing PyCharm failed with the following crash:

```
piet:~$ umake ide pycharm
Choose installation path: /home/piet/tools/ide/pycharm
Downloading and installing requirements                                                                                                              |
ERROR: Unhandled exception########################################################################################################################## |
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/umake/tools.py", line 154, in wrapper
    function(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/umake/frameworks/baseinstaller.py", line 262, in get_progress
    self.pbar.update(progress)
  File "/usr/lib/python3/dist-packages/progressbar/__init__.py", line 271, in update
    raise ValueError('Value out of range')
ValueError: Value out of range
piet:~$
```

However, re-running `umake ide pycharm` completes the installation successfully.

I can only offer some speculation about the cause:

1. I *might* have resized the terminal window during the first run, so my first guess was that the progressbar library might have a SIGWINCH-related bug in its range calculation. However, glancing at the code, it doesn't look like its `self.maxval` gets touched by SIGWINCH, so this seems unlikely.
2. My next guess is that `BaseInstaller.get_progress()`'s tricky calculation of the progress value might sometimes be overshooting the default `maxval`: perhaps due to some edge case in updating the download progress values, or even due to float rounding?

Package versions:

* python3-apt: 0.9.3.11build1
* python3-argcomplete: 0.8.1-0ubuntu1
* python3-bs4: 4.3.2-2ubuntu2
* python3-gi: 3.14.0-1
* python3-pexpect: 3.2-1
* python3-progressbar: 2.3-1
* python3-requests: 2.4.3-6
* python3-xdg: 0.25-4
* python3-yaml: 3.11-2
* python3: 3.4.3-1
* ubuntu-make: 0.6.2

pjdelport:
#75 seems to be either the same bug, or closely related.

didrocks:
Hey pjdelport, sorry for the late reply, I was on holidays until today :)

[...]

PR #133, opened 2015-08-06
#123 Fix progress bar out of range exceptions

syndbg:
Addresses #123 and it's the same occurring issue in #121 .

This is the simplest fix I could think of.

### DIFF
--- a/umake/frameworks/baseinstaller.py
+++ b/umake/frameworks/baseinstaller.py
@@ -255,10 +255,16 @@ class BaseInstaller(umake.frameworks.BaseFramework):
                                                             (self.pkg_size_download + self.total_download_size),
                                                             0.15)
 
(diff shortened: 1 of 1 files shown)


didrocks:
Thanks for the fix. I guess you couldn't get a nice reproducer on this either, right?
Maybe you can add some logger.debug() with the various values when it exceeds 100% before you normalize it (that way, as we run some tests in debug mode, we would see which values are incorrect)?

Anyway, it's useful to safeguard it, just rename the function and I'll merge it! Big thanks!

syndbg:
@didrocks Addressed the method visibility.

The only way I (could) reproduce it is in #121 when I try to install a non-SE Eclipse IDE.

didrocks:
Good! Merging :)

If you have it when installing a non-SE Eclipse IDE, can you try to add some prints to the involved variables and results? I'm interested :)

Thanks!

PR #134, opened 2015-08-11
Fix umake exit status for failed downloads

### LINKED ISSUE #129: umake exits with 0 if download page has changed or unavailable
It appears that currently when umake exits due to the download page being unavailable or having changed, it exits with an exit status of 0 – it shouldn't be 0 after exiting due to an error. This is currently the default behavior in BaseInstaller.

Also perhaps it would be a good idea to also check whether umake's exit status is 0 after a successful install in our large tests.

oijazsh:
Fixes #129 - umake should now exit with the correct status when it fails. All tests now confirm umake's exit code except for those in which umake is terminated by sendcontrol('C'). All but two medium tests pass; the two that fail are "Reinstall android studio on another path (non empty) once installed should remove the first version" and "We prompt if we try to install on an existing directory which isn't empty" but these were already failing for me before my changes.

I only added one test per framework that overloads get_metadata_and_license() although some of them have more than one exit point e.g. visual studio code can also fail if the License page is changed. If needed I can add tests for those as well.

I had to adjust the arduino installer in order to exit it properly for the medium test. We had to confirm soup.find() had returned something other than None before accessing ['href'].

### DIFF
--- a/tests/large/test_web.py
+++ b/tests/large/test_web.py
@@ -66,7 +66,7 @@ class FirefoxDevTests(LargeFrameworkTests):
         self.child = pexpect.spawnu(self.command('{} web firefox-dev'.format(UMAKE)))
         self.expect_and_no_warn("Firefox Dev is already installed.*\[.*\] ")
         self.child.sendline()
-        self.wait_and_no_warn()
+        self.wait_and_close()
 
     def test_default_install(self):
         """Install firefox dev from scratch test case"""
@@ -77,7 +77,7 @@ class FirefoxDevTests(LargeFrameworkTests):
         self.expect_and_no_warn("Choose language:")
         self.child.sendline("")
         self.expect_and_no_warn("Installation done", timeout=self.TIMEOUT_INSTALL_PROGRESS)
-        self.wait_and_no_warn()
+        self.wait_and_close()
         self.verify_install(install_language)
 
     def test_arg_language_select_install(self):
@@ -87,7 +87,7 @@ class FirefoxDevTests(LargeFrameworkTests):
         self.expect_and_no_warn("Choose installation path: {}".format(self.installed_path))
         self.child.sendline("")
         self.expect_and_no_warn("Installation done", timeout=self.TIMEOUT_INSTALL_PROGRESS)
-        self.wait_and_no_warn()
+        self.wait_and_close()
         self.verify_install(install_language)
 
     def test_interactive_language_select_install(self):
@@ -99,16 +99,18 @@ class FirefoxDevTests(LargeFrameworkTests):
         self.expect_and_no_warn("Choose language:")
         self.child.sendline(install_language)
         self.expect_and_no_warn("Installation done", timeout=self.TIMEOUT_INSTALL_PROGRESS)
-        self.wait_and_no_warn()
+        self.wait_and_close()
         self.verify_install(install_language)
 
     def test_unavailable_language_select_install(self):
+        """Installing Firefox-dev in unavailable language should be rejected"""
         install_language = "ABCdwXYZ"
         self.child = pexpect.spawnu(self.command('{} web firefox-dev --lang={}'.format(UMAKE, install_language)))
         self.expect_and_no_warn("Choose installation path: {}".format(self.installed_path))
-        self.accept_default_and_wait(expect_warn=True)
-        self.child.close()
-        self.assertEqual(self.child.exitstatus, 1)
+        self.child.sendline("")
+        self.wait_and_close(expect_warn=True, exit_status=1)
+
+        self.assertFalse(self.launcher_exists_and_is_pinned(self.desktop_filename))
 
     def language_file_exists(self, language):
         return self.path_exists(os.path.join(self.installed_path, "dictionaries", "{}.aff".format(language)))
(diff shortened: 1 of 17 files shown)


didrocks:
Hard and excellent work, I indeed added the status_code later in the process (as return_main_screen() was mostly implemented in case we add a graphical UI one day, in addition to the cli UI) but didn't realize I wasn't paying attention to the exit code at all! This pull request is more than appreciated (and good catch on the soup.find() gotcha).

All your changes looks good to me. I have run the medium tests with and without your changes, and they all pass for me (are the failures reproducible for you? Would be interesting if you can paste the error you are seeing for investigation).

I have a small comment on just one possible factorization for medium tests, tell me what you think, otherwise the additional helpers and changes are +1 for me :)
Having the additional test for the license page changing makes sense as well, if you have the willing to implement it.

Thanks again!

oijazsh:
I will happy to reduce the duplication by writing the tests to use a helper method.

The issue I see with the second idea is that this test will not work properly with certain frameworks such as the former firefox-dev installation which did not use a download page. As such making a single test_invalid_path in ContainerTests could be problematic. We could perhaps work around that perhaps by raising a SkipTest if download_file_path is None.

If its OK, I'll go ahead with adding a method for the new test to ContainerTests. Let me know what you think.

Oh and about the failed tests, I recently had to recover my home dir from backup and they do not seem to be failing anymore somehow.

Also, sorry for the late reply - I was terribly ill till the past week.

didrocks:
On the late reply -> no worry, I was on a more positive "probably on holidays" option for you! I hope that everything is fine at least now. From experience, I normally wait for a month without any activity to know if people are still interested into working on it FYI ;)

Please go ahead with the helper method path, it seems the most logical after you answered and your preferred option as well. Sounds a little bit less magical than the one with raising a SkipTest for download_page being None.

I have even less clue for the medium test restore then :p However, happy that they pass for you!
Keep me posted if you need anything, thanks agian.

oijazsh:
Change committed - let me know if its good now.

Note that some files in server content had CRLF line endings When writing back during our tests these are replaced with LF line endings. It should not be an issue - just an extra commit should be required after the tests are run for the first time.

Oh and all medium tests currently pass.

didrocks:
Do you mind rebasing on trunk (and check for the new frameworks)? seems there are some conflicts to resolve with yesterday's addition I guess. If you don't have the time for it, I can have look.

oijazsh:
Rebased and adjusted for the two new frameworks.

tests/large/test_web.py
```diff
@@ -99,16 +99,18 @@ def test_interactive_language_select_install(self):
         self.expect_and_no_warn("Choose language:")
         self.child.sendline(install_language)
         self.expect_and_no_warn("Installation done", timeout=self.TIMEOUT_INSTALL_PROGRESS)
-        self.wait_and_no_warn()
+        self.wait_and_close()
         self.verify_install(install_language)
 
     def test_unavailable_language_select_install(self):
+        """Installing Firefox-dev in unavailable language should be rejected"""
```
didrocks: Thanks for adding the description :)

tests/large/test_web.py
```diff
@@ -99,16 +99,18 @@ def test_interactive_language_select_install(self):
...
         install_language = "ABCdwXYZ"
         self.child = pexpect.spawnu(self.command('{} web firefox-dev --lang={}'.format(UMAKE, install_language)))
         self.expect_and_no_warn("Choose installation path: {}".format(self.installed_path))
-        self.accept_default_and_wait(expect_warn=True)
-        self.child.close()
-        self.assertEqual(self.child.exitstatus, 1)
+        self.child.sendline("")
+        self.wait_and_close(expect_warn=True, exit_status=1)
+
+        self.assertFalse(self.launcher_exists_and_is_pinned(self.desktop_filename))
```
didrocks: Good catch! Thanks for adding the additional assert ;) (I wonder if in the future, we shouldn't factor this in an key_files_installed(), but that will be for later anyway)

tests/medium/__init__.py
```diff
@@ -177,3 +180,14 @@ def create_file(self, path, content):
         command = self.command_as_list(["mkdir", "-p", dir_path, path])
         if not self._exec_command(command)[0]:
             raise BaseException("Couldn't create {} in container".format(path))
+
+    @nottest
```
didrocks: Why do you decorate this call (as it doesn't start with test_)?

tests/medium/test_ide.py
```diff
@@ -22,7 +22,9 @@
 
 from . import ContainerTests
 import os
+
```
didrocks: good catch as well, thanks for cleaning up for consistency ;)

didrocks:
Everything looks good to me and ready to merge! I just have this @notest question, and then, I'll be able to push the trigger :)

Really good work for this tedious tasks, and thanks for the additional cleanups as well!

tests/medium/__init__.py
oijazsh: It has test on a word boundary which appears to be good enough for nose. I was surprised too:

http://nose.readthedocs.org/en/latest/writing_tests.html

didrocks: Waow! We learn something everyday :)

didrocks:
With this rationale, time to merge your excellent changes!

Thanks again for looking into this, much appreciated and raising the overall quality ;)
I plan to cut next release next Tuesday if everything goes right.

PR #135, opened 2015-08-12
Fix Visual Studio Code download URL

### LINKED ISSUE #136: VSCode: ERROR: Download page changed its syntax or is not parsable
Microsoft just released an update to VSCode, and with it, changed the way that download URLs work on their site.

As a result, the script no longer works, erroring in

```text
ERROR: Download page changed its syntax or is not parsable
```

I've gone ahead and created a PR (#135) which fixes this issue.

vsimonian:
VSCode updated their download mechanism on their website.

This code fetches the download URL using the new method.

Fixes #136


didrocks:
Thanks a lot for those changes and spotting it just a few hours after the breakage!

Your change looks good, I'm going to refresh the medium tests assets to match thoes new urls and tags and then, will do a release :)
Nice work!

vsimonian:
Glad it all worked out. Thanks!

PR #142, opened 2015-08-25
Add support for installing the Android SDK only

sschuberth:


didrocks:
This looks good for a first attempt, well done! :)

I wonder about few minor things to finish this work and get it merged:
-  do you think the "tools" path should be added in the user's PATH? That way, when they type adb or the server is starting, those are those android sdk tools that are used. You can see in the Go framework some implementation of that (with the corresponding tests as well).
- Isn't there any end user license agreement for the SDK (like there are in Android Studio) when you download it from the web page? I didn't check, so unsure. If there is, I guess the android studio framework + corresponding tests adaptation would be enough.
- Finally, the large tests look good, I guess you run it, right? `./runtests tests/large/test_android.py:AndroidSDKTests` to only run those . However, I doubt you tested the medium one `./runtests test/medium/test_android.py:AndroidSDKContainer` as you didn't provide some fake assets? Fortunately, the certificates and so on are already present for developer.android.com, so you only need to ship them in tests/data/server-content/developer.android.com/ and it should be quite straightforward.

Do not hesitate if you need any help, I have one finale remark inlined :)
Nice work!

sschuberth:
> do you think the "tools" path should be added in the user's PATH?

Yes. In fact both `tools` and `platform-tools` usually are required. I'll add those. However, I did not find any tests in the Go framework specific to adding directories to PATH.

> Isn't there any end user license agreement for the SDK

Yes, there is. But there's also one for the *N*DK, and as I was mostly copying code from there I though this would already be handled somehow.

> Finally, the large tests look good, I guess you run it, right?

I indeed did run the large tests, but I did run the medium tests in a wrong way (I used `./runtests test/medium/test_android.py:test_android.AndroidSDKTests` which succeeded).

I'll soon force-push an update.

sschuberth:
I just pushed a version that addresses most of your comment and asst fake assets. However, when running the medium tests now I still get

```
ERROR: Install android sdk from scratch test case
  File "/home/sschuberth/Development/ubuntu-make/tests/large/test_android.py", line 313, in test_default_android_sdk_install
    stderr=subprocess.DEVNULL),
  File "/usr/lib/python3.4/subprocess.py", line 561, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['sshpass', '-p', 'user', 'ssh', '-o', 'UserKnownHostsFile=/dev/null', '-o', 'StrictHostKeyChecking=no', '-t', '-q', 'user@172.17.0.47', "/home/sschuberth/Development/ubuntu-make/tests/tools/run_in_umake_dir /umake '/home/user/tools/android/android-sdk/tools/android list'"]' returned non-zero exit status 127
```

Looks to me as if some command line argument quoting if wrong to pass the `list` argument to `android`, or?

sschuberth:
@didrocks Would you have some time to help me make the medium test pass?

didrocks:
Sorry, I was certain that I answered you before the week-end to your first answer.

1) I saw that you found and push how to set PATH user env variable, great!
2) I think you still have the license thing to implement, isn't it? Would you like some help with this? (I guess you can get your inspiration from Android Studio framework itself). Extra bonus if you add it to the NDK, but I can do it myself as well, if you don't feel that much :)
3) IIRC, one of my inline comment (maybe I didn't post it and my browser crashed) was to add a test to ensure android and another tools are in the PATH in the tests. You just need something like line 68 in the test_go framework: https://github.com/ubuntu/ubuntu-make/blob/master/tests/large/test_go.py#L68 (probably doubling it to check an exec in both paths). Of course, the mock assets needs to have those in their archives. For the second tool you will check the path for, ensure you use os.path.join(self.installed_path, "…", "…") in the test_* function, as self.installed_path is replaced in the medium tests by the correct path in your setUp() (and so, after calling parent's setUp()), don't get trapped by this! :)
- I looked at it last week and I think you are right, I bet it's trying to exec("/home/user/tools/android/android-sdk/tools/android list") as a single arg[0], and so, don't find it (which is what the exit status of 127 backs up)

I'm looking at the last point in a couple of hours, all the rest looks good. Do you mind working on the 2) and 3) (3 should really take less than 5 minutes).

Keeping you posted on fixing the exec in medium test one!

sschuberth:
> I think you still have the license thing to implement, isn't it?

probably, yes. I was a bit confused as I did not find such logic for the NDK yet, although it also displays a license page, and that the download also seemed to work without the license thing. I'll try to have a look at it anyway.

> add a test to ensure android and another tools are in the PATH in the tests

Done. Thanks for the hint to the line in `test.go.py`.

I'll force-push an update in a second.

didrocks:
I have fixed the Android NDK to ask for a license. Even if we can bypass the license acceptance, I don't think it's appropriate to do so, that's why everytime upstream required a license to be acked, we present them to the user.

I did some refactoring so that you can reuse the license parser from Android Studio (and now the NDK!) and fixed a bug when the license appears after the download links in the web page.

So, please rebase on master first to get those fixes. Then, you just need to mirror something like https://github.com/ubuntu/ubuntu-make/commit/145ddafc1eaf605a1e503c48a47a878604ab6d98 to require and parse the license (it should now be shown when you execute interactively).
Then for large and medium tests, just look at https://github.com/ubuntu/ubuntu-make/commit/8abcc7aa845455d4850900561d9189540a558532. There is just a couple of lines to add for the large tests. However, as the container doesn't have UTF8 support, to avoid raising an exception, I just remove the UTF8 charsets in the mock assets for now.

As you are using the same page than for android studio and the sdk which is presented is exactly the same than the android studio one, I think you just need to parse_license() from android studio, with the same tag + add the large tests couple of lines.

Hope this helps ;)

I'm now looking at the medium container argument issue.

sschuberth:
Thanks. I've rebased onto your master. I also fixed the way I was calling `add_env_to_user`: I'm now making two separate calls instead of passing PATH twice in the map, which did not work as keys need to be unique. Still, the newly added second PATH test is failing for some reason.

I'll first look into adding the license stuff now.

didrocks:
Ah, something I just note as I'm trying to run the medium tests: you didn't change
tests/data/server-content/developer.android.com/sdk/index.html to point to your local mock (so, as in my case, the docker containers are blocking any Internet connection, I was able to spot it).

* You need to change <a onclick="return onDownload(this)" id="linux-tools" href="http://dl.google.com/android/android-sdk_r24.0.1-linux.tgz">android-sdk_r24.0.1-linux.tg
z</a>
to point to your fake version (look for android studio url, where I use https://developer.android.com/android-studio-fake.tgz) :)
* Also, you didn't update the checksum to match your fake asset. Once you fix the url, it will fail downloading it.
* Finally, your matching url download was a little bit too large. The one I used in NDK was fine as it's the only downloadable item ni the page. On the SDK page, you have both Android Studio bundle and Android SDK being offered for download, so you need something more restrictive like "linux-tools"

The good news is that here is the diff which is fixing the 3 points above: https://gist.github.com/didrocks/ad37ef883b4ff6bed462

With this, there is no more issue of "android list" not working (so Ubuntu Make copes well in that case), I wonder though if that just means that android SDK has some download requirements that are not matched. Do you think we should install openjdk for instance?

sschuberth:
Thanks for all for help. I've incorporated your patches / addressed the comments and force-pushed a new version. With this medium tests pass for me, but large test do not. For some reason the test for the second path added to PATH fails:

	FAIL: Install android sdk from scratch test case
	  File "/home/seschube/Development/ubuntu-make/tests/large/test_android.py", line 312, in test_default_android_sdk_install
	    self.assertTrue(self.is_in_path(os.path.join(self.installed_path, "platform-tools", "adb")))
	AssertionError: False is not true

Any ideas?

sschuberth:
> Do you think we should install openjdk for instance?

You certainly need a JRE to run the `android` tool as its a Java application. So yes, we should probably depend on Open JDK (or Oracle Java). Do you mean that we'd need to add that dependency to the Docker container, or that it should get installed on the user's system when `umake` installs the Android SDK?

didrocks:
-> Just checked on the PATH thingy, here is what I've found:

I don't find any adb or any platform-tools when download the Android SDK. So the tests works and say that adb isn't in the PATH :)
There is though a "platforms" directory, but that one is empty as it will contains downloaded platforms version.
So, the good news is that the tests works, adb isn't part of the shipped SDK (maybe you need to change what you add to the PATH to only add "tools/"

-> On dependencies:
yeah, so let's make the dependencies matching what we do for Android Studio (as we have some tools here depending on openjdk and IIRC, the android sdk tools need some i386 binaries).
This means add:
packages_requirements=["openjdk-7-jdk", "libncurses5:i386", "libstdc++6:i386", "zlib1g:i386",
                                                "jayatana"],
as in Android Studio (I think jayatana is still a good thing to have as some sdk tools are graphical and appmenu integration is interesting)

and for the medium tests, readd the fake repo:
self.apt_repo_override_path = os.path.join(self.APT_FAKE_REPO_PATH, 'android') just after self.port.

I think then, if you are happy with the result, that we can merge! :)

sschuberth:
> I don't find any adb or any platform-tools when download the Android SDK.

Doh, you're right. The user is supposed to manually install the platform-tools via the SDK manager once it has been installed. So I'll remove that test, but I'll keep the code that adds the `platform-tools` directory to `PATH` as user usually rely on being able to call `adb`.

> let's make the dependencies matching what we do for Android Studio

Done.

Please have a final (hopefully :-) look at the PR!

didrocks:
All looks good, agreed with your "let's add platforms-tools to the path now so that it can be used later on".

One last thing though! :p

I guess there is a bug in Ubuntu Make when you call multiple times the same PATH thing, I only see the latest one being added in ~/.profile. I'll fix it in the future, however, I think though that rather than having this:
+        add_env_to_user(self.name, {"PATH": {"value": os.path.join(self.install_path, "tools")}})
+        # ensure "adb" is in PATH once the plaform-tools are installed via the SDK manager
+        add_env_to_user(self.name, {"PATH": {"value": os.path.join(self.install_path, "platform-tools")}})

I would just rather have one call to add_env_to_user, like: (/!\ untested!)
add_env_to_user(self.name, {"PATH": {"value": "{}:{}".format(os.path.join(self.install_path, "tools"), os.path.join(self.install_path, "platform-tools"))}})
What do you think?

It seems that you didn't update to remove the "platform-tools" directory from the mock assets, mind doing so? (I prefer to have the mock reflecting as much as possible the real thing). Do not forget to change the sha1sum in index.html :)

And finally, what do you think about changing the finale message to tell to run the SDK manager? (even maybe with the exact path) to install the platform-tools?

My strategy would be:
- do a release tomorrow for unity3D editor support (finishing working on this this afternoon)
- then merge you PR for scheduling another release next week (I don't want your excellent work being overlooked by the press if I announce Unity3d editor support within the same release, hence the delay).
Does this sound good to you?

Thanks again for working on this :)

sschuberth:
> I guess there is a bug in Ubuntu Make when you call multiple times the same PATH thing

Are you sure that you're using my latest code? At first I had

    add_env_to_user(self.name, {"PATH": {"value": os.path.join(self.install_path, "tools")},
                                                    "PATH": {"value": os.path.join(self.install_path, "platform-tools")}}})

which indeed indeed causes the issue you're seeing, as `PATH` is used as the key in a map, and you cannot add keys twice, so the value of the key is overwritten instead.

However, with the two separate calls to `add_env_to_user()` my tests showed that it worked.

> It seems that you didn't update to remove the "platform-tools" directory from the mock assets, mind doing so?

Done.

> And finally, what do you think about changing the finale message to tell to run the SDK manager?

Not sure about this. The platform-tools are just one of several packages that you need to install to get started, and they are checked by default if you start the SDK manager. I think just pointing at the web page for adding packages is fine.

I did another force-push.

didrocks:
> Are you sure that you're using my latest code?
I just deleted your branch, and fetching it again. I installed it with umake and get in ~/.profile only:
# Ubuntu make installation of Android SDK
PATH=/home/didrocks/tools/android/android-sdk/platform-tools:$PATH

I guess the issue for me is that adb is already in my path (I have other packages installed), and so, it skips it. I'm going to open an Ubuntu Make bug as it should ensure the proposed path is the same. So, don't worry about it, on a vanilla system, it works well (and that's why it works on medium tests)

>I think just pointing at the web page for adding packages is fine.
Ack then :)
Just change it from UI.display to UI.delayed_display. Otherwise, you can get:
Downloading and installing requirements                                                                                                               |
100% |################################################################################################################################################|
Installing Android SDK
SDK installed in /home/didrocks/tools/android/android-sdk. More information on how to use it on https://developer.android.com/sdk/installing/adding-packages.html
|#####################################################################################################################################################|
You need to restart a shell session for your installation to work

(so, ugly ####### duplication)

Then force push, and we are ready to merge tomorrow after the release!

sschuberth:
> I guess the issue for me is that adb is already in my path

Wait a second, `adb` is in the `platform-tools`, to according to your theory it should be the `tools` directory that's being skipped, not `platform-tools`, or?

> Just change it from UI.display to UI.delayed_display.

Done.

didrocks:
> Wait a second, adb is in the platform-tools, so according to your theory it should be the platform-tools directory that's being skipped, not tools, or?

Sorry, yeah, I was talking about the "android" binary :)
I reread my code for add_to_user_env and my theory is wrong, it really needs to have one call per env variable as I'm calling:
remove_framework_envs_from_user(framework_tag) before readding it. Consequently, only latest PATH addition "platform-tools" would work.

And that makes sense for future upgrade support and such.
So, yeah, the way to fix this is either:
to support multiple keys for one elem, like:
add_env_to_user(self.name, {"PATH": [os.path.join(self.install_path, "tools"),  os.path.join(self.install_path, "platform-tools")]})
I need to change some Ubuntu Make code for supporting this, but this is doable easily.

or concat it yourself as I proposed (this works without changing anything in Make):
add_env_to_user(self.name, {"PATH": {"value": "{}:{}".format(os.path.join(self.install_path, "tools"), os.path.join(self.install_path, "platform-tools"))}})

What's the format you prefer? (just push one and I'll adapt Ubuntu make trunk to make it work with it if it doesn't)

sschuberth:
I'd prefer the first syntax, allowing the value to be a list. Did you deliberately omit the literal `"value"` key from your first example? I think that's a nice simplification.

I also prefer to not just push the use of this without it being implemented, but wait for you to do so and then rebase & fixup again.

didrocks:
> I'd prefer the first syntax, allowing the value to be a list. Did you deliberately omit the literal "value" key from your first example? I think that's a nice simplification.

That was on purpose, before a read the docline and the tests I added almost a year ago ;) We have a dictionary because the "keep" parameter is for some specific case we had in the past where we want to overwrite or append to an existing env variable (that way, we can control this through another dictionary value, "keep").

So, I just implemented the tuple element in this commit (pushed to master):
https://github.com/ubuntu/ubuntu-make/commit/e66e0d31baf4ebd3bd44f0202a620177e396ee6f

I did rerun all the small tests to ensure it doesn't regress anything else (and so, other frameworks), You can base your work on this and then, we should be good to go! :)

sschuberth:
Wow, this is odd. So I did the changes in the `WIP` commit 95bbce7, and now suddenly the medium tests break again. I thought I should be able to get something as easy as this done, but apparently I'm not :-(

didrocks:
I just push the Unity3D editor support to trunk. I can have a look tomorrow morning on why the medium tests are broken, I know it's quite difficult to debug them when you don't really know how the docker implementation has been done.

Do not hesitate! :)

sschuberth:
Note sure why, but after rebasing onto your Unity3D changes the medium tests suddenly worked again. Maybe cd6b930 did the trick, as I was `PermissionError: [Errno 1] Operation not permitted`errors on the temporary `.tgz` archive.

didrocks:
Ah, you had that issue? Interesting, it's one that I see sometimes on the large tests in the datacenter but never reproduced it here. I have no access to those machines and could never get interesting logs either.
It started like 4 months ago.

If you get it again at any point, please ping me, I would be interested in knowing what's happening (my bet is the apt database being in a locked state and so not accepting installing some packages).

So, I guess your branch is good to pull just after I cut next release tomorrow morning! Nice work! Thanks again for being so reponsive, working with you on this was a pleasure :)

sschuberth:
> If you get it again at any point, please ping me

Will do, probably by mail then.

> Thanks again for being so reponsive, working with you on this has been a pleasure :)

Likewise! It's nice to contribute to a project where one finally gets quick review cycles.

didrocks:
Now that Ubuntu Make 15.09 is release, it's time to push this "Merge pull request" button :) Thanks again!
(will be released next Tuesday, probably)

sschuberth:
Nice to see this merged, and thanks for your help!

PR #114, opened 2015-06-20
Add Scala support

ivuk:
This is not complete yet, as it's lacking tests.

I was just wondering if something like this would be suitable for inclusion? It's basically a two-way merge between existing Dart and Go frameworks, with minor alterations. :)

As far as the tests are concerned, I've intended to reuse the existing medium and large Go tests, since they seem to be applicable for Scala as well.

Also, a disclaimer - I've been considering picking up Scala, so I'm no great expert at it; to me, this seems similar enough to what `apt-get install scala` provides. :)

PR #143, opened 2015-08-25
Netbeans support

fcole90:
Initial Netbeans support

### DIFF
--- a/tests/large/test_ide.py
+++ b/tests/large/test_ide.py
@@ -449,3 +449,44 @@ class ArduinoIDETests(LargeFrameworkTests):
         self.expect_and_no_warn("Arduino is already installed.*\[.*\] ")
         self.child.sendline()
         self.wait_and_close()
+
+
+class BaseNetBeansTests(LargeFrameworkTests):
+    """Tests for the Netbeans installer."""
+
+    TIMEOUT_INSTALL_PROGRESS = 120
+    TIMEOUT_START = 60
+    TIMEOUT_STOP = 60
+
+    def setUp(self):
+        super().setUp()
+        self.installed_path = os.path.expanduser("~/tools/ide/netbeans")
+        self.desktop_filename = "netbeans.desktop"
+
+    def test_default_install(self):
+        """Install from scratch test case"""
+        self.child = spawn_process(self.command('{} ide netbeans'.format(UMAKE)))
+        self.expect_and_no_warn("Choose installation path: {}".format(self.installed_path))
+        self.child.sendline("")
+        self.expect_and_no_warn("Installation done", timeout=self.TIMEOUT_INSTALL_PROGRESS)
+        self.wait_and_close()
+
+        logger.info("Installed, running...")
+
+        # we have an installed launcher, added to the launcher and an icon file
+        self.assertTrue(self.launcher_exists_and_is_pinned(self.desktop_filename))
+        self.assert_exec_exists()
+        self.assert_icon_exists()
+
+        # launch it, send SIGTERM and check that it exits fine
+        proc = subprocess.Popen(self.command_as_list(self.exec_path), stdout=subprocess.DEVNULL,
+                                stderr=subprocess.DEVNULL)
+
+        self.check_and_kill_process(["java", self.installed_path], wait_before=self.TIMEOUT_START)
+        proc.wait(self.TIMEOUT_STOP)
+
+        # ensure that it's detected as installed:
+        self.child = spawn_process(self.command('{} ide netbeans'.format(UMAKE)))
+        self.expect_and_no_warn("Netbeans is already installed.*\[.*\] ")
+        self.child.sendline()
+        self.wait_and_close()
--- a/umake/frameworks/ide.py
+++ b/umake/frameworks/ide.py
@@ -41,7 +41,6 @@ from umake.network.download_center import DownloadCenter, DownloadItem
 from umake.tools import create_launcher, get_application_desktop_file, ChecksumType, Checksum, MainLoop
 from umake.ui import UI
 
-
 logger = logging.getLogger(__name__)
 
 
@@ -466,3 +465,112 @@ class Arduino(umake.frameworks.baseinstaller.BaseInstaller):
                                                      categories=categories))
         if not self.was_in_arduino_group:
             UI.delayed_display(DisplayMessage(_("You need to logout and login again for your installation to work")))
+
+
+class BaseNetBeans(umake.frameworks.baseinstaller.BaseInstaller):
+    """The base for all Netbeans installers."""
+
+    BASE_URL = "http://download.netbeans.org/netbeans/"
+    FALLBACK_VERSION = "8.0.2"
+    EXECUTABLE = "nb/netbeans"
+
+    def __init__(self, category, flavour=""):
+        """The constructor.
+        @param category The IDE category.
+        @param flavour The Netbeans flavour (plugins bundled).
+        """
+        # add a separator to the string, like -cpp
+        if flavour:
+            flavour = '-' + flavour
+        self.flavour = flavour
+
+        super().__init__(name="Netbeans",
+                         description=_("Netbeans IDE"),
+                         category=category,
+                         only_on_archs=['i386', 'amd64'],
+                         download_page="https://netbeans.org/downloads/zip.html",
+                         dir_to_decompress_in_tarball="*",
+                         desktop_filename="netbeans{}.desktop".format(flavour),
+                         packages_requirements=['openjdk-7-jdk', 'jayatana'])
+
+    @MainLoop.in_mainloop_thread
+    def get_metadata_and_check_license(self, result):
+        """Get the latest version and trigger the download of the download_page file.
+        :param result: the file downloaded by DownloadCenter, contains a web page
+        """
+        # Processing the string to obtain metadata (version)
+        try:
+            url_version_str = result[self.download_page].buffer.read().decode('utf-8')
+        except AttributeError:
+            # The file could not be parsed or there is no network connection
+            logger.error("You either have no connection available or "
+                         "the download page changed its syntax or is not parsable")
+            UI.return_main_screen(status_code=1)
+
+        preg = re.compile(".*/images_www/v6/download/.*")
+        for line in url_version_str.split("\n"):
+            if preg.match(line):
+                line = line.replace("var PAGE_ARTIFACTS_LOCATION = \"/images"
+                                    "_www/v6/download/", "").replace("/\";", "")
+                self.version = line.strip()
+
+        if not self.version:
+            # Fallback
+            logger.warning("Could not determine latest version, using version "
+                           "{} as fallback...".format(self.FALLBACK_VERSION))
+            self.version = self.FALLBACK_VERSION
+
+        self.version_download_page = "https://netbeans.org/images_www/v6/download/" \
+                                     "{}/js/files.js".format(self.version)
+        DownloadCenter([DownloadItem(self.version_download_page)], self.parse_download_page_callback, download=False)
+
+    @MainLoop.in_mainloop_thread
+    def parse_download_page_callback(self, result):
+        """Get the download_url and trigger the download and installation of the app.
+        :param result: the file downloaded by DownloadCenter, contains js functions with download urls
+        """
+        logger.info("Netbeans {}".format(self.version))
+
+        # Processing the string to obtain metadata (download url)
+        try:
+            url_file = result[self.version_download_page].buffer.read().decode('utf-8')
+        except AttributeError:
+            # The file could not be parsed
+            logger.error("The download page changed its syntax or is not parsable")
+            UI.return_main_screen(status_code=1)
+
+        preg = re.compile('add_file\("zip/netbeans-' + self.version +
+                          '-[0-9]{12}' + self.flavour + '.zip"')
+        for line in url_file.split("\n"):
+            if preg.match(line):
+                # Clean up the string from js (it's a function call)
+                line = line.replace("add_file(", "").replace(");", "").replace('"', "")
+                url_string = line
+
+        if not url_string:
+            # The file could not be parsed
+            logger.error("The download page changed its syntax or is not parsable")
+            UI.return_main_screen(status_code=1)
+
+        string_array = url_string.split(", ")
+        try:
+            url_suffix = string_array[0]
+            md5 = string_array[2]
+        except IndexError:
+            # The file could not be parsed
+            logger.error("The download page changed its syntax or is not parsable")
+            UI.return_main_screen(status_code=1)
+
+        self.complete_name = "netbeans{}".format(self.flavour)
+        download_url = "{}{}/final/{}".format(self.BASE_URL, self.version, url_suffix)
+        self.download_requests.append(DownloadItem(download_url, Checksum(ChecksumType.md5, md5)))
+        self.start_download_and_install()
+
+    def post_install(self):
+        """Create the Netbeans launcher"""
+        create_launcher(self.desktop_filename,
+                        get_application_desktop_file(name=_("Netbeans IDE"),
+                                                     icon_path=join(self.install_path, "nb", "netbeans.png"),
+                                                     exec=join(self.install_path, "bin", "netbeans"),
+                                                     comment=_("Netbeans IDE"),
+                                                     categories="Development;IDE;"))
--- a/tests/medium/test_ide.py
+++ b/tests/medium/test_ide.py
@@ -226,3 +226,19 @@ class ArduinoIDEInContainer(ContainerTests, test_ide.ArduinoIDETests):
         umake_command = self.command('{} ide arduino'.format(UMAKE))
         self.bad_download_page_test(umake_command, download_page_file_path)
         self.assertFalse(self.launcher_exists_and_is_pinned(self.desktop_filename))
+
+
+class BaseNetBeansInContainer(ContainerTests, test_ide.BaseNetBeansTests):
+    """This will test the NetBeans IDE integration inside a container"""
+
+    TIMEOUT_START = 20
+    TIMEOUT_STOP = 10
+
+    def setUp(self):
+        self.hostnames = ["www.netbeans.org"]
+        self.port = "443"
+        # Reuse the Android Studio environment.
+        self.apt_repo_override_path = os.path.join(self.APT_FAKE_REPO_PATH, 'android')
+        super().setUp()
+        # override with container path
+        self.installed_path = os.path.join(self.install_base_path, "ide", "netbeans")


didrocks:
This looks like a nice first draft, well done and thanks for looking into this ;)

I have quite some remarks, some minor (in term of style coherence), some more important (like the versions which should be abstract from umake I guess and also, avoiding any urlopen in __init__() while keeping the UI reactive).

Nice work on thise nevertheless, look at my inline comments! Do not hesitate if you need help/suggestions on any of this. Once this is settled down, we can look at the tests :)

umake/frameworks/ide.py
```diff
@@ -473,3 +474,83 @@ def post_install(self):
                                                      categories=categories))
         if not self.was_in_arduino_group:
             UI.delayed_display(DisplayMessage(_("You need to logout and login again for your installation to work")))
+            
+class BaseNetBeans(umake.frameworks.baseinstaller.BaseInstaller):
+    """The base for all Netbeans installers."""
+    
+    BASE_URL = "http://download.netbeans.org/netbeans/8.0.2/final/"
```
didrocks: I would prefer we do not encode any version here, as the goal of Ubuntu Make is to always be able to download latest versions of any IDE, and back that with tests. Is there a way to go to a more generic download page, get the version (or a redirect, that is transparently supported by the framework) which will lead to the desired page?

umake/frameworks/ide.py
```diff
@@ -473,3 +474,83 @@ def post_install(self):
...
+
+    def __init__(self, category, version="8.0.2", flavour=""):
```
didrocks: So, same here, I would really like to remove any kind of "version", especially in description, and so on. We always pull latest, whatever it is at the current moment (I know that eclipse Luna is an exception, but I want to remove this, see some other pull requests :))

umake/frameworks/ide.py
```diff
@@ -473,3 +474,83 @@ def post_install(self):
...
+        #add a separator to the string, like -cpp
+        if flavour != "":
+            flavour = '-' + flavour
+        super().__init__(name="Netbeans",
+                         description=_("Netbeans IDE " + version),
+                         category=category, 
+                         only_on_archs=['i386', 'amd64', 'arm'],
```
didrocks: "arm" doesn't exist on ubuntu, I guess you meant armhf (or maybe armel or both?)

umake/frameworks/ide.py
```diff
@@ -473,3 +474,83 @@ def post_install(self):
...
+                         download_page=None,
+                         dir_to_decompress_in_tarball="netbeans" + flavour,
+                         desktop_filename="netbeans" + flavour + ".desktop",
```
didrocks: I would prefer that you follow how we mostly do string manipulation in umake, that would be in those cases:
dir_to_decompress_in_tarball="netbeans{}".format(flavour),
desktop_filename="netbeans{}.desktop".format(flavour),

Please follow the same pattern in other string manipulation below, just for the sake of consistency :)

I'm not an english native, but I think it's more "flavor" than "flavour", right?
Also, we support wildcard for dir decompression, so, if there is only one directory in the tarball, you can use something like:
dir_to_decompress_in_tarball="netbeans*"

umake/frameworks/ide.py
```diff
@@ -473,3 +474,83 @@ def post_install(self):
...
+        self.url_string = self.search_url_string(self, version, flavour)
+        data = self.url_string.split(", ")
+        self.url_suffix = data[0]
+        self.md5sum = data[2]
+        self.complete_name = "netbeans" + flavour
+    
+    @property
+    def search_url_string(self, version_number, flavour=""):
+      #connect to the page and look for the version
+      data = urlopen("https://netbeans.org/images_www/v6/download/" \
```
didrocks: Please do not use urlopen() from anything that is in in __init__(). This would mean that everytime you:
umake --help
or even:
umake <tab><tab>, you are going to run that request. I think you should only do that in the first step of installation (so, from download_provider_page) and download it asynchronously through the DownloadCenter (to not block the UI on long request).

umake/frameworks/ide.py
```diff
@@ -473,3 +474,83 @@ def post_install(self):
...
+                line = line.replace("add_file(", "").replace(");", "")
+                line = line.replace('"', "")
+                
+                return line
+    '''        
+    @property
+    @abstractmethod
+    def executable(self):
+        pass
+    '''
```
didrocks: I guess you can now remove those comments?

umake/frameworks/ide.py
fcole90: Yes, I can cut the string after "netbeans" and make the version a variable

umake/frameworks/ide.py
fcole90: Ok, but that might be a bit more difficult, but I'll try to find a way

umake/frameworks/ide.py
fcole90: I don't know, there's a folder called Linux-arm, I'll investigate on that too

umake/frameworks/ide.py
fcole90: I will update to follow that style, it's the first time that I work only using the source code as documentation 

In UK it's flavour, but I can change it if that makes things more consistent.

I think I misunderstood the meaning of the dir_to_decompress_in_tarball variable. I thought it was the destination folder. Then I will need to change it, as the required dirs are almost all but those with Mac OS and Windows specific libraries.

umake/frameworks/ide.py
fcole90: Ok, it will be updated soon

umake/frameworks/ide.py
fcole90: Yes, I did it already in my branch.

didrocks:
Excellent! Having a look tomorrow morning first thing ;)

didrocks:
That's great progress, nice that you can get the latest version by parsing the page! I just had a first look, as there is some structural changes I think it will be better to put them in the general comment section, then, I'll go on some minor details in the PR itself.

My guess (as your FIXME comment said) is to use the framework as most as possible. Indeed, your urlopen() requests are sync, while the downloads provided by the framework are aysnc.

So, here is how I would do it:
1. Setting download_page="https://netbeans.org/downloads/zip.html" in __init__()
2. Overriding:
    @MainLoop.in_mainloop_thread

This method is the callback of the main download, running in the mainloop thread (so, remove download_page_provider override). Here, you proceed the remaining work that was initially in your get_latest_version(), knowing that the downloaded page content is at result[self.download_page].buffer

3. Then, you build a new DownloadItem (for the javascript page with the correct version url). Like:
self.version_download_page = https://netbeans.org/images_www/v6/download/" \
+                           "{}/js/files.js".format(self.version)
DownloadCenter([DownloadItem(self.version_download_page)], self.parse_download_page_callback, download=False)

4. in your new parse_download_page_callback(self, result) callback (that you must decorate with @MainLoop.in_mainloop_thread):
result[self.version_download_page].buffer will contain then information ready for parsing and downloading. So with this, I would put in that function both what you were doing in your download_provider_page() and search_url_string(). You finish by:
self.download_requests.append(DownloadItem(download_url, Checksum(ChecksumType.md5, md5)))
-> this should download and do the md5sum check for you, (so no need for on_done callback and such). I'm unsure about your comment that it tried to download the md5 url instead of checking it, we can have a look after this refactoring, that makes more sense I guess, doesn't it?

And then, you should be good, of course, do not forget to check if anything doesn't match the parsing you expected (as you override some BaseInstaller methods doing this for you) to warn the user. You can have a look at framework/BaseInstaller.py for the error message and just copy them. Something like:
logger.error("Download page changed its syntax or is not parsable")
UI.return_main_screen()
(this stops the execution of the method by raising an exception, so no need for return after it)

Does it sound cray? I think you will be more async and aligned with the framework this way, even if you do this "2 downloads jump" (sad that we can't infer easily with a link "last version", the DownloadCenter even support multiple downloads at once!), I guess the user experience will be better ;)

Do not hesitate if you need any help/precision. I'm now going to put some minor style comments that can be reused in the refactoring.

Oh, last thing, in case you didn't notice, you can run easily the pep8 style checking for the project with: ./runtests pep8.

Thanks again for your contribution, keep me posted!

umake/frameworks/ide.py
```diff
@@ -473,3 +474,115 @@ def post_install(self):
...
+    
+    BASE_URL = "http://download.netbeans.org/netbeans/"
+    executable = "nb/netbeans"
+
+    def __init__(self, category, flavour=""):
+        """ The constructror.
+        @param category The IDE category.
+        @param flavour The Netbeans flavour (plugins bundled).
+        """
+        #add a separator to the string, like -cpp
```
didrocks: Please add a space between # and start of the comment (if I didn't do it everywhere, blame me :p)

umake/frameworks/ide.py
```diff
@@ -473,3 +474,115 @@ def post_install(self):
...
+        if flavour != "":
```
didrocks: I prefer the more pythonic "if flavour:"

umake/frameworks/ide.py
```diff
@@ -473,3 +474,115 @@ def post_install(self):
...
+            flavour = '-' + flavour
+        self.flavour = flavour
+        
+        super().__init__(name="Netbeans",
+                         description=_("Netbeans IDE"),
+                         category=category, 
+                         only_on_archs=['i386', 'amd64'], #FIXME add arm support
+                         download_page=None,
+                         dir_to_decompress_in_tarball="*", #FIXME windows and mac files not required
```
didrocks: Are all linux/windows/mac stuff in different directories? In that case, you can use something like:
dir_to_decompress_in_tarball="linux-*", (wildcards are supported as you saw)

And the decompressor will do it for you.

umake/frameworks/ide.py
```diff
@@ -473,3 +474,115 @@ def post_install(self):
...
+                         desktop_filename="netbeans{}.desktop".format(flavour),
+                         #FIXME support for oracle java and multiple java versions
```
didrocks: This is a general "problem" through Ubuntu Make itself. However, I tend to only get one java version. For Oracle vs OpenJDK, I still wait for Canonical's lawer to give me feedbacks if we can download it (presenting the license) legally. It seems to be quite a complex problem (remember what happened to the package in the ubuntu repository which was doing it).

So, I would say "wait and see"

umake/frameworks/ide.py
```diff
@@ -473,3 +474,115 @@ def post_install(self):
...
+        @return A string in the form: suffix, 8digit number, md5sum, languages
+        """
+        preg = re.compile('add_file\("zip/netbeans-' + version\
+                        + '-[0-9]{12}' + flavour + '.zip"')
+        for line in url_file:
+            for line in url_file.split("\n"):
+                if preg.match(line):
+                    #Clean up the string from js (it's a function call)
+                    line = line.replace("add_file(", "").replace(");", "")
+                    line = line.replace('"', "")
```
didrocks: You can chain it once more with the upper line ;)

umake/frameworks/ide.py
```diff
@@ -473,3 +474,115 @@ def post_install(self):
...
+        url_version_str = url_version.read().decode('utf-8')
+        preg = re.compile(".*/images_www/v6/download/.*")
+        for line in url_version_str.split("\n"):
+            if preg.match(line):
+                line = line.replace("var PAGE_ARTIFACTS_LOCATION = \"/images" + \
+                                    "_www/v6/download/", "").replace("/\";", "")
+                return line.strip()
+        #Fallback
+        logger.error("Could not determine latest version, using fallback...")
+        return "8.0.2"
```
didrocks: I would prefer such hardcoded fallbacks to be a static property of the class on top, to ensure we don't have to read through the class implementation to change it.

So something next to BASE_URL, like FALLBACK_VERSION="8.0.2"

umake/frameworks/ide.py
```diff
@@ -473,3 +474,115 @@ def post_install(self):
...
+        url_version = urlopen("https://netbeans.org/downloads/zip.html")
+        url_version_str = url_version.read().decode('utf-8')
+        preg = re.compile(".*/images_www/v6/download/.*")
+        for line in url_version_str.split("\n"):
+            if preg.match(line):
+                line = line.replace("var PAGE_ARTIFACTS_LOCATION = \"/images" + \
+                                    "_www/v6/download/", "").replace("/\";", "")
+                return line.strip()
+        #Fallback
+        logger.error("Could not determine latest version, using fallback...")
```
didrocks: We recover from the error, so a warning rather?

umake/frameworks/ide.py
```diff
@@ -473,3 +474,115 @@ def post_install(self):
                                                      categories=categories))
         if not self.was_in_arduino_group:
             UI.delayed_display(DisplayMessage(_("You need to logout and login again for your installation to work")))
+            
+class BaseNetBeans(umake.frameworks.baseinstaller.BaseInstaller):
+    """The base for all Netbeans installers."""
+    
+    BASE_URL = "http://download.netbeans.org/netbeans/"
+    executable = "nb/netbeans"
```
didrocks: I prefer capital letters for any static members (easier to spot in the method call)

umake/frameworks/ide.py
```diff
@@ -473,3 +474,115 @@ def post_install(self):
...
+
+    def __init__(self, category, flavour=""):
+        """ The constructror.
```
didrocks: small typo: construcror -> constructor :)

umake/frameworks/ide.py
```diff
@@ -473,3 +474,115 @@ def post_install(self):
...
+                           "{}/js/files.js".format(self.version))
+        #Extract the file from the response                 
+        url_file_str = url_file.read().decode('utf-8')
+        url_string = self.search_url_string(url_file_str, self.version, self.flavour)
+        string_array = url_string.split(", ")
+        url_suffix = string_array[0]
+        md5 = string_array[2]
+        self.complete_name = "netbeans{}".format(self.flavour)
+
+        download_url = "{}{}{}{}".format(self.BASE_URL, self.version, "/final/", url_suffix)
```
didrocks: guess "/final" can be in the string to format, so:
"{}{}/final/{}"

umake/frameworks/ide.py
```diff
@@ -473,3 +474,115 @@ def post_install(self):
...
+        
+    def post_install(self):
+        """Create the Netbeans launcher"""
+        icon_path = self.install_path + 'nb/netbeans'
+        exec_path = self.install_path + 'bin/netbeans'
+        comment = _("Netbeans IDE " + self.version)
+        categories = "Development;IDE;"
+        create_launcher(self.desktop_filename,
+                        get_application_desktop_file(name=_("Netbeans IDE " \
+                                                     + self.version),
```
didrocks: No version in the name itself please. If the framework can later autoupdate, it's something we don't want to support (especially if it's upstream's autoupdating tool). The promise we give is to always download latest stable version.

umake/frameworks/ide.py
```diff
@@ -473,3 +474,115 @@ def post_install(self):
...
+        if not isfile(join(self.install_path, self.executable)):
+            logger.debug("{} binary isn't installed".format(self.name))
+            return False
+        return True
+        
+    def post_install(self):
+        """Create the Netbeans launcher"""
+        icon_path = self.install_path + 'nb/netbeans'
+        exec_path = self.install_path + 'bin/netbeans'
+        comment = _("Netbeans IDE " + self.version)
```
didrocks: Same comment than 2 lines below on versions ;) (and yeah, see the other PR, I want to remove "Luna" for eclipse which is the only counter-example)

umake/frameworks/ide.py
```diff
@@ -473,3 +474,115 @@ def post_install(self):
...
+        categories = "Development;IDE;"
```
didrocks: I know that some other fameworks in ide.py are doing this (setting all variables and then calling create_launcher). I do prefer the inline style though and will clean up (or maybe you want to do it in another PR?)

umake/frameworks/ide.py
```diff
@@ -473,3 +474,115 @@ def post_install(self):
...
+        if not super().is_installed:
+            return False
+        if not isfile(join(self.install_path, self.executable)):
+            logger.debug("{} binary isn't installed".format(self.name))
+            return False
+        return True
+        
+    def post_install(self):
+        """Create the Netbeans launcher"""
+        icon_path = self.install_path + 'nb/netbeans'
```
didrocks: Nitpick: please use join(), like join(self.install_path, "nb", "netbeans")

fcole90:
Thank you for your comments, as soon as I have some free time I will look better into ;)

umake/frameworks/ide.py
fcole90: No, most of the directories are common to every system, so we would need a wildcard for the folders NOT to select, as they're just few.

fcole90:
Hi, I'm back here after some time. I completed the tasks you asked me, and polished my code a bit. Let me know if there's any required improvement ;)

didrocks:
Hey! Excellent, I hope you have succeeded in your exams!

So, things look good in principle, do you mind rebasing on master and trying to squash your commits in logical chunks so that the diffs are easier to read? That will help me doing a proper review :)

Then, we can work on the large and medium tests together :)

fcole90:
Thank you! I will rebase on master soon (possibly this weekend). About the commits, is it a problem making one single chunk with something like "Added support for Netbeans with possible extension to Netbeans flavours"? Should I make a new pull request for that?

didrocks:
Yeah, as you are only touching one file and one place, a single chunk (after your rebase on master) sounds good. Just do an interactive rebase and squash your commits in one: https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history

One thing: you don't need the is_installed() property anymore, just look at other frameworks in master, we only need an executable path now :)

Then, if you have time, have a look at the large tests (tests/large/test_ide.py) for other frameworks. You can:
- run pep8 tests with ./runtests pep8
- run large tests for one class with ./runtests tests/large/test_ide.py:<Your_New_Test_Class>

I hope this helps! Do not hesitate if you have any questions

fcole90:
Thank you for your suggestions, I'm still learning :D Let me know if it needs any other imporvement ;)
It passes the PEP8 test and the small one, but not the medium one.
The following is the error that comes from the medium test:
```
ERROR: Failure: ValueError (No such test BaseNetbeans)
  File "/usr/lib/python3/dist-packages/nose/failure.py", line 42, in runTest
    raise self.exc_class(self.exc_val)
ValueError: No such test BaseNetbeans
```
I'm running the large test while I'm writing, I'm updating as soon as it finishes.

didrocks:
No worry on learning! Thanks for squashing the commits ;)

I have some comments on the PR itself (just did add those).

I didn't see your commit for your large test, so I don't know why it's failing, you should just copy and adapt what's in tests/large/test_ide.py for instance for one framework, that will give us a base to discuss!

Keep me posted about your changes, we're getting there ;)

fcole90:
I added the necessary test (I didn't understand I had to write it, until you told me that I thought there was a generic test that checked any IDE class). Currently the test is correctly downloading and installing the app but fails in checking its execution:

```
======================================================================
ERROR: Install from scratch test case
  File "/home/fabio/development/python/ubuntu-make/tests/large/test_ide.py", line 492, in test_default_install
    self.check_and_kill_process(["java", self.installed_path], wait_before=self.TIMEOUT_START)
  File "/home/fabio/development/python/ubuntu-make/tests/large/__init__.py", line 83, in check_and_kill_process
    pid = self._pid_for(process_grep)
  File "/home/fabio/development/python/ubuntu-make/tests/large/__init__.py", line 78, in _pid_for
    raise BaseException("The process that we can find with {} isn't started".format(process_grep))
nose.proxy.BaseException: The process that we can find with ['java', '/home/fabio/tools/ide/netbeans-test'] isn't started
-------------------- >> begin captured logging << --------------------
2015-10-15 21:50:34,967 [tests.large.test_ide] INFO: Installed, running...
--------------------- >> end captured logging << ---------------------

Ran 1 test in 60.408s
```
I'm wondering if that's fault of my procedure to enable jayatana. Soon I will try again removing that string ;)

didrocks:
What you need to check here is the netbean process, to have a way to ensure that the netbean IDE is started and running.

So, I would say, look at your processes when you run netbean on your machine (ps aux), and see what process string you can find here.
_pid_for will do something like ps aux | grep 'first string' | grep 'second string'… so that you only have one result, the netbean process. Making sense?

fcole90:
I was expecting something like that, so I think I should be able to see the NetBeans splash screen when it gets launched (which doesn't happen). I tried a clean install of netbeans with umake and it seems to work (I can launch it from the launcher created by umake). I just checked from System Monitor and ps aux and its process is just named java, so it should get killed correctly. I'm investigating ;)

didrocks:
Look at the whole ps aux line output, to ensure we don't match a wrong process, we check that all of the parameters set (including the path, thus), is present

fcole90:
Yes, look at this:
```
ps aux | grep java | grep ~/tools/ide/netbeans/
fabio     1796  0.0  0.0  16912  3476 ?        S    14:31   0:00 /bin/bash /home/fabio/tools/ide/netbeans/platform/lib/nbexec --userdir /home/fabio/.netbeans/8.0.2 --cachedir /home/fabio/.cache/netbeans/8.0.2 --jdkhome  --branding nb --clusters /home/fabio/tools/ide/netbeans/nb:/home/fabio/tools/ide/netbeans/ergonomics:/home/fabio/tools/ide/netbeans/ide:/home/fabio/tools/ide/netbeans/extide:/home/fabio/tools/ide/netbeans/java:/home/fabio/tools/ide/netbeans/apisupport:/home/fabio/tools/ide/netbeans/webcommon:/home/fabio/tools/ide/netbeans/websvccommon:/home/fabio/tools/ide/netbeans/enterprise:/home/fabio/tools/ide/netbeans/mobility:/home/fabio/tools/ide/netbeans/profiler:/home/fabio/tools/ide/netbeans/python:/home/fabio/tools/ide/netbeans/php:/home/fabio/tools/ide/netbeans/identity:/home/fabio/tools/ide/netbeans/harness:/home/fabio/tools/ide/netbeans/cnd:/home/fabio/tools/ide/netbeans/dlight:/home/fabio/tools/ide/netbeans/groovy:/home/fabio/tools/ide/netbeans/extra:/home/fabio/tools/ide/netbeans/javacard:/home/fabio/tools/ide/netbeans/javafx: -J-Dnetbeans.importclass=org.netbeans.upgrade.AutoUpgrade -J-Dnetbeans.accept_license_class=org.netbeans.license.AcceptLicense -L-XX:MaxPermSize=384m -J-Xmx1024m -J-client -J-Xss2m -J-Xms32m -J-XX:PermSize=32m -J-Dapple.laf.useScreenMenuBar=true -J-Dapple.awt.graphics.UseQuartz=true -J-Dsun.java2d.noddraw=true -J-Dsun.java2d.dpiaware=true -J-Dsun.zip.disableMemoryMapping=true
fabio     2082  126  7.9 6401380 640236 ?      Sl   14:31   0:41 /usr/lib/jvm/java-8-oracle/bin/java -Djdk.home=/usr/lib/jvm/java-8-oracle -classpath /home/fabio/tools/ide/netbeans/platform/lib/boot.jar:/home/fabio/tools/ide/netbeans/platform/lib/org-openide-modules.jar:/home/fabio/tools/ide/netbeans/platform/lib/org-openide-util.jar:/home/fabio/tools/ide/netbeans/platform/lib/org-openide-util-lookup.jar:/home/fabio/tools/ide/netbeans/platform/lib/locale/boot_ja.jar:/home/fabio/tools/ide/netbeans/platform/lib/locale/boot_pt_BR.jar:/home/fabio/tools/ide/netbeans/platform/lib/locale/boot_ru.jar:/home/fabio/tools/ide/netbeans/platform/lib/locale/boot_zh_CN.jar:/home/fabio/tools/ide/netbeans/platform/lib/locale/org-openide-modules_ja.jar:/home/fabio/tools/ide/netbeans/platform/lib/locale/org-openide-modules_pt_BR.jar:/home/fabio/tools/ide/netbeans/platform/lib/locale/org-openide-modules_ru.jar:/home/fabio/tools/ide/netbeans/platform/lib/locale/org-openide-modules_zh_CN.jar:/home/fabio/tools/ide/netbeans/platform/lib/locale/org-openide-util_ja.jar:/home/fabio/tools/ide/netbeans/platform/lib/locale/org-openide-util-lookup_ja.jar:/home/fabio/tools/ide/netbeans/platform/lib/locale/org-openide-util-lookup_pt_BR.jar:/home/fabio/tools/ide/netbeans/platform/lib/locale/org-openide-util-lookup_ru.jar:/home/fabio/tools/ide/netbeans/platform/lib/locale/org-openide-util-lookup_zh_CN.jar:/home/fabio/tools/ide/netbeans/platform/lib/locale/org-openide-util_pt_BR.jar:/home/fabio/tools/ide/netbeans/platform/lib/locale/org-openide-util_ru.jar:/home/fabio/tools/ide/netbeans/platform/lib/locale/org-openide-util_zh_CN.jar:/usr/lib/jvm/java-8-oracle/lib/dt.jar:/usr/lib/jvm/java-8-oracle/lib/tools.jar -Dnetbeans.default_userdir_root=/home/fabio/.netbeans -Dnetbeans.running.environment=gnome -Dnetbeans.dirs=/home/fabio/tools/ide/netbeans/nb:/home/fabio/tools/ide/netbeans/ergonomics:/home/fabio/tools/ide/netbeans/ide:/home/fabio/tools/ide/netbeans/extide:/home/fabio/tools/ide/netbeans/java:/home/fabio/tools/ide/netbeans/apisupport:/home/fabio/tools/ide/netbeans/webcommon:/home/fabio/tools/ide/netbeans/websvccommon:/home/fabio/tools/ide/netbeans/enterprise:/home/fabio/tools/ide/netbeans/mobility:/home/fabio/tools/ide/netbeans/profiler:/home/fabio/tools/ide/netbeans/python:/home/fabio/tools/ide/netbeans/php:/home/fabio/tools/ide/netbeans/identity:/home/fabio/tools/ide/netbeans/harness:/home/fabio/tools/ide/netbeans/cnd:/home/fabio/tools/ide/netbeans/dlight:/home/fabio/tools/ide/netbeans/groovy:/home/fabio/tools/ide/netbeans/extra:/home/fabio/tools/ide/netbeans/javacard:/home/fabio/tools/ide/netbeans/javafx: -Dnetbeans.home=/home/fabio/tools/ide/netbeans/platform -Dnetbeans.importclass=org.netbeans.upgrade.AutoUpgrade -Dnetbeans.accept_license_class=org.netbeans.license.AcceptLicense -Xmx1024m -client -Xss2m -Xms32m -XX:PermSize=32m -Dapple.laf.useScreenMenuBar=true -Dapple.awt.graphics.UseQuartz=true -Dsun.java2d.noddraw=true -Dsun.java2d.dpiaware=true -Dsun.zip.disableMemoryMapping=true -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/home/fabio/.netbeans/8.0.2/var/log/heapdump.hprof -splash:/home/fabio/.cache/netbeans/8.0.2/splash.png org.netbeans.Main --cachedir /home/fabio/.cache/netbeans/8.0.2 --userdir /home/fabio/.netbeans/8.0.2 --branding nb
```
Those two are the processes that get spawned when I launch NetBeans

fcole90:
If I ```killall java``` they disappear and appear again launching Netbeans.

didrocks:
The first one is the parent of the second one it seems by its pid number, maybe try to kill that one (and then, it won't relaunching Netbeans)

fcole90:
It gets closed everytime any of the two is closed. Is it normal that when it gets closed it even deletes the ```~/.local/share/applications/netbeans.desktop``` file?

didrocks:
yeah, the tests are deleting everything to be idempotent. You can just put a return on top of the setUp() function in tests/large/__init__.py

fcole90:
Yes, it's ok for the tests (still your return trick will be useful in some occasions). But the problem is that it gets deleted even when I manually launch netbeans and then kill it :open_mouth:

didrocks:
urgh, seriously? Seems to be on the netbean side then? Maybe contact upstream to know why you get that behavior…

fcole90:
Probably, I'm still checking what's happening..

 I need to ask you few more questions:
  1. Do you know if it is possible to simulate the download of the zip file and instead just extract and install it? I'm wasting some time on that because it takes a lot to download it.
  2. umake is installing netbeans, it creates even its file .desktop, but doesn't pin it on the launcher (not always at least, is it nondeterministic?). Do you know how can I debug that?
  3. I think that netbeans doesn't get killed by the test because it doesn't get launched. Is it possible to check a difference between those conditions?

Thank you very much for your help!

didrocks:
1. Sure! That's the goal of medium tests, but it's quite hard for someone not knowing the system how to get it working. I would suggest:
a) in the framework code switch from https to http and to localhost (http://localhost:8000) and same for BASE_URL (so that you only have one local server to setup)
b) create a netbeans/ directory somewhere, and download the real assets there, respecting the same directory structure than on the server
c) change the url download path inside the .html file you downloaded to reference localhost
d) just run python3 -m http.server while being in that directory, then you will have your local server!
I guess then, the turnaround time to run the test will be way faster.

2. That's not an issue on our side, it's a known glib bug (which should be better now in vivid/wily than trusty). Are you testing there?

3. yeah, just before the check_and_kill_process() call, just add something like:
sleep(10000)
print(self._pid_for([<the array of string with java, blablabla>]))
if a valid pid is printed, then, netbean is detected.

didrocks:
and thank *you* for working on this ;)

fcole90:
1. Sounds great! I'll try and let you know.
 
 2. Yes, but I have a Wily partition too, if needed I can use it or comment that line when using Trusty.

 3. Perfect!

Thanks, I'll let you know soon ;)

fcole90:
Seems it was not being launched, removing the jayatana string solved the issue and it's now passing tests. Maybe I can remove that, for the moment, to complete the current task and add the jayatana additional support in due time. Let me know ;)

didrocks:
Yeah, the reason why I disabled it by default is that it was crashing some apps, maybe netbeans is one of them… so keep it disabled by default at least for now.

fcole90:
Ok, I'm cleaning things up and squashing commits again ;)

I don't think that in this case it's a problem with jayatana itself (as it works when launching NetBeans manually) but instead it has to do with the way  enable it. The .desktop file recognizes that syntax, but maybe it's not the same for the method that's calling it during the tests.

fcole90:
I rebased on master and squelched the commits (I messed things for a moment but seems to be again OK :sweat_smile:). Let me know! :)

tests/large/test_ide.py
```diff
@@ -450,3 +450,50 @@ def test_default_install(self):
         self.expect_and_no_warn("Arduino is already installed.*\[.*\] ")
         self.child.sendline()
         self.wait_and_close()
+
+
+class BaseNetBeansTests(LargeFrameworkTests):
+    """The base for all Netbeans installers."""
```
didrocks: I guess you want to change the description for the tests? :)

tests/large/test_ide.py
```diff
@@ -450,3 +450,50 @@ def test_default_install(self):
...
+
+    TIMEOUT_INSTALL_PROGRESS = 600
```
didrocks: Please put back the timeout to 120. The timeout is not the full install time, it's the time when nothing is print/changing on the CLI. So basically: no  download progressing, or anything. If you don't get 0.5-1% of download during those 120 minutes, it means there is a big issue :)

tests/large/test_ide.py
```diff
@@ -450,3 +450,50 @@ def test_default_install(self):
...
+    TIMEOUT_START = 60
+    TIMEOUT_STOP = 60
+
+    def setUp(self):
+
```
didrocks: extra line to remove

tests/large/test_ide.py
```diff
@@ -450,3 +450,50 @@ def test_default_install(self):
...
+        super().setUp()
+        self.installed_path = os.path.expanduser("~/tools/ide/netbeans")
+        self.desktop_filename = "netbeans.desktop"
+
+    @property
+    def arch_option(self):
```
didrocks: You don't use that property, does it means that 32 and 64 bits are the same?

If so, please remove that property to avoid dead test code.

umake/frameworks/ide.py
```diff
@@ -466,3 +465,113 @@ def post_install(self):
...
+                url_string = line
+
+        if not url_string:
+            # The file could not be parsed
+            logger.error("The download page changed its syntax or is not parsable")
+            UI.return_main_screen(status_code=1)
+
+        string_array = url_string.split(", ")
+        try:
+
```
didrocks: Uneeded line space, just delete it

didrocks:
Everything looks good, and I've tested it both manually and running the tests, nice work!
In addition to the cosmetic comments from before, I would add that you can squash in 2 commits instead of 1 (but no hard requirement here):
- 1 for the code itself
- 1 for the large tests

I would do it like that, but no pressure :)

So remains the medium tests. This is a little bit more complex than the large one. I'm happy to handle it once you have done the above change if you don't have the time.
Basically the idea here is that the tests are running in a docker container, with fake assets and fake ssl certificates (and so, you need to rebuild the docker image to pick that certificate). tests/medium/test_android.py is a good example for this.

I would ensure they are 3 tests in it:
- the normal one (just provided with fake assets) and inheritence
- having one with download page changing its syntax and thus failing
- another one using the fallback system

There may have some issues as you are using the same domain, but oncee over https and once over http, so on 2 differents ports that our testing frameworks don't support yet. Hence my willingness to give a hand on this this week :)
Keep me posted! Once again, good work!

fcole90:
Thank you, and thanks for your great help! :)

I did the changes you asked me and rebased on top of master. If I have some time I will have a look at the medium test in the weekend ;)

didrocks:
All looks good, see my small comment ;)

Keep me posted on the medium tests and do not hesitate if you need a hand, I know it's not the most easy things to tests when you don't know the internals.

fcole90:
Spawned commit ;)

fcole90:
The following is the error that I get running any medium test (tried with PhpStorm, Arduino and NetBeans):
```
./runtests tests/medium/test_ide.py:BaseNetBeansInContainer
[nose.plugins.manager] DEBUG: DefaultPluginManager load plugin nose_json = nose_json.plugin:JsonReportPlugin
[nose.plugins.manager] DEBUG: DefaultPluginManager load plugin cov = nose_cov:Cov
[nose_cov] DEBUG: nose-cov options
[nose_cov] DEBUG: nose-cov configure
E
======================================================================
ERROR: Install from scratch test case
  File "/home/fabio/development/python/ubuntu-make/tests/medium/test_ide.py", line 242, in setUp
    super().setUp()
  File "/home/fabio/development/python/ubuntu-make/tests/medium/__init__.py", line 79, in setUp
    self.container_id = subprocess.check_output(command).decode("utf-8").strip()
  File "/usr/lib/python3.4/subprocess.py", line 607, in check_output
    with Popen(*popenargs, stdout=PIPE, **kwargs) as process:
  File "/usr/lib/python3.4/subprocess.py", line 859, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.4/subprocess.py", line 1379, in _execute_child
    executable = os.fsencode(executable)
  File "/usr/lib/python3.4/os.py", line 770, in fsencode
    raise TypeError("expect bytes or str, not %s" % type(filename).__name__)
TypeError: expect bytes or str, not NoneType

Ran 1 test in 0.001s

FAILED (errors=1)
```
I don't know how to debug that :/

fcole90:
I tried both the large and medium tests again in wily but with the same result

didrocks:
Interesting, I doubt though the issue in the large tests are the same. (see my answer about pexpect and dimensions).
For medium tests: did you install docker and can you try starting any docker image? It can't get any container id from what was started.

fcole90:
About the medium test, can you point me a few more basics things on how should I work, and what the test does? :)

didrocks:
First, try to run an existing one of them successfully (like ./runtests tests/medium/test_dart.py).
If anything is failing, please ensure you have docker install and that you can run a docker container successfully.

Then, do as for the other medium tests: create a new class inheriting from others (your large tests + the container class), and try to mimick what other frameworks are doing. Do not create any test_* method (as we inherit from your large tests, you will already have one: test_default_install().

The "complex" part is to create fake assets (the medium tests are using the docker container to download from there). Check what I wrote in both https://github.com/ubuntu/ubuntu-make/pull/114#issuecomment-122133562 and in https://github.com/ubuntu/ubuntu-make/pull/114#issuecomment-127552858, which should give you a start :)

Again, I know that part isn't the most straightforward to get, and as soon as ubuntu 15.10 is out, I'm more than happy to give a hand on this :)

fcole90:
Thank you very much, I'm having a look at those things and letting you know ;)

fcole90:
Hi, I tried multiple time to run the medium tests (trusty) but with no success. Doker is installed, I downloaded an ubuntu image (didrocks/umake, something like that) and tried to run medium test inside it (after installing umake dependencies) but without success. I'm updating the comment with the log as soon as I have the time to run the tests again ;)

didrocks:
I just rerun all medium tests from master to ensure that things are working as expected and they all pass.
Ensure that you have python3-pexpect 4.0.1-0~ppatrusty6 and python3-ptyprocess 0.5-1~trusty1 from
https://launchpad.net/~didrocks/+archive/ubuntu/ubuntu-developer-tools-center/+packages and you should be good to go :)

didrocks:
Ok, https://github.com/ubuntu/ubuntu-make/commit/e58d47b123f81d8527928e707679ce1b61704dfd should help you btw. That way, you can have your medium tests providing multiple servers on different ports (you have 80 and 443 in your case): netbeans.org and download.netbeans.org. You should rebase on master to get that functionality ;)

I'm available for any help if you want me to finish the medium test work, I know it can be quite complex. Do not hesitate! ;)

fcole90:
I think it's better if I leave the medium test to you. It requires some knowledge that I can't

didrocks:
Sure! You already did some excellent jobs, thanks for this!

I'll try to handle this on Monday/Tuesday, I'll ping you back to that you have an idea of what it was and maybe it will help you implementing more netbeans support? :)
Thanks again for your tremendous work here!

fcole90:
I took a bit more of time to check the things but the medium test seems to require more knowledge than what I currently have and I'm not able to fill the gap for the moment. If you could finish it I would be very happy, thank you.

fcole90:
Thank you very, much, glad to hear that :) Thank you again ;)

didrocks:
And here we go! Sorry for the additional day, I wanted to cut a release first :)

I've made some small changes starting from https://github.com/ubuntu/ubuntu-make/commit/637d133a8a584dd1981e75207b6e6adde0e85ced to https://github.com/ubuntu/ubuntu-make/commit/0427713fdf84a7341ffc80a96472e06b2bfdc38d:
- make the code a little bit more compliant with other parts of the code on this project
- add the medium assets + certificates + dockerfile to push a new docker image with the additional server
- changed the medium test server port/name to server on those
- as a bonus (last commit) added some more tests for various failures!

I hope they are understandable by themselves, if you want any further explanation, do not hesitate! Thanks again for your work here and if you want to implement any other netbeans flavours, now that the base is there, you are more than welcomed! :)

fcole90:
Thank you, that's awesome :) As soon as I have some free time I'll implement some flavours too ;)

Issue #81, opened 2015-04-05
Install developer tools to $XDG_DATA_HOME as default.


marcinant:
I think that it could be nice to keep home directory clean and install developer tools to ~/.local/share folder. Binaries could go to ~/.local/bin.
This path could be also added to $PATH with this code in ~/.profile
```
if [ -d "$HOME/.local/bin" ] ; then
    PATH="$HOME/.local/bin:$PATH"
fi
```

In this way applications could be available to run from command line.

didrocks:
Hi marcinant,

Thanks for the suggestion! I guess installing applications in other places than ~/tools makes sense.
Some things to note though:
- XDG_DATA_HOME isn't supposed to contain softwares, but user's data. From the xdg spec (http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html): "There is a single base directory relative to which user-specific data files should be written. This directory is defined by the environment variable $XDG_DATA_HOME."
- I don't want to separate binaries from the software package (in an eventual .local/bin).  That breaks most of the auto-update system by the usptream softwares.
There is a tool to add to user's $PATH on a per framework basics in ubuntu-make. If you feel that some are missing, we can work on this (we don't add them if there is a .desktop file pointing to that binary usually).

If we can find a better default path for installing softwares (and compatible with the XDG specs) though, I'm opened to the data. Any suggestions?

marcinant:
Well, if you want to be so attached to the standards and recommendations then in fact `umake` should not install any software in userspace. Everything should go to: /opt or /usr/local. However we all know it's not the point as `umake` is for an easy rootless installation and **updates**.

[FHS](http://www.pathname.com/fhs/pub/fhs-2.3.html#HOMEUSERHOMEDIRECTORIES) says:
>/home is a fairly standard concept, but it is clearly a site-specific filesystem. [9] The setup will differ >from host to host. **Therefore, no program should rely on this location.**

There are documents such as [Ubuntu HomeFolder](https://help.ubuntu.com/community/HomeFolder#Installing_Software_Into_The_Home_Directory)
or [Arch filesystem hierarchy](https://wiki.archlinux.org/index.php/Arch_filesystem_hierarchy#.2Fhome:_User_home_directories):
> These are the user home directories, which can be found under '/home/$USER' (~/). Within their home directory, a user can write files, delete them, **install programs**, etc.

However there is also [this](http://standards.freedesktop.org/menu-spec/latest/apcs02.html) document which says:

[...]

PR #154, opened 2015-10-01
Install dir

oijazsh:
This PR changes default install directory to $XDG_DATA_HOME/umake (#81). All paths used in tests have been changed to ~/.local/share/umake and a separate test has been added to confirm installation with $XDG_DATA_HOME set - similar to the existing test custom install path. 

For the new test, I could not think of a more elegant way to set an environment variable in a running container so I simply used /bin/bash -c "export $XDG_DATA_HOME && /bin/umake"'. This required changing the run_in_umake_dir script to not strip the quotations - and then make changes accordingly.

I should mention that I have not been able to rerun all of the large tests due a rather slow Internet connection.

### DIFF
--- a/tests/large/__init__.py
+++ b/tests/large/__init__.py
@@ -28,7 +28,7 @@ import stat
 import subprocess
 from time import sleep
 from umake.tools import get_icon_path, get_launcher_path, launcher_exists_and_is_pinned, remove_framework_envs_from_user
-from ..tools import LoggedTestCase, get_path_from_desktop_file, is_in_group
+from ..tools import LoggedTestCase, get_path_from_desktop_file, is_in_group, INSTALL_DIR
 
 
 class LargeFrameworkTests(LoggedTestCase):
@@ -40,7 +40,7 @@ class LargeFrameworkTests(LoggedTestCase):
         self.installed_path = ""
         self.framework_name_for_profile = ""
         self.conf_path = os.path.expanduser("~/.config/umake")
-        self.install_base_path = os.path.expanduser("~/.local/share/umake")
+        self.install_base_path = os.path.expanduser("~/{}".format(INSTALL_DIR))
         self.desktop_filename = ""
         self.child = None
         self.additional_dirs = []
--- a/tests/large/__init__.py
+++ b/tests/large/__init__.py
@@ -40,6 +40,7 @@ class LargeFrameworkTests(LoggedTestCase):
         self.installed_path = ""
         self.framework_name_for_profile = ""
         self.conf_path = os.path.expanduser("~/.config/umake")
+        self.install_base_path = os.path.expanduser("~/.local/share/umake")
         self.desktop_filename = ""
         self.child = None
         self.additional_dirs = []
--- a/tests/large/test_android.py
+++ b/tests/large/test_android.py
@@ -216,6 +216,18 @@ class AndroidStudioTests(LargeFrameworkTests):
             self.assertTrue(self.launcher_exists_and_is_pinned(self.desktop_filename))
             self.assert_exec_exists()
 
+    def test_xdg_data_install_path(self):
+        """Install in path specified by XDG_DATA_HOME"""
+        xdg_data_path = "/tmp/foo"
+        self.installed_path = "{}/umake/android/android-studio".format(xdg_data_path)
+        self.child = pexpect.spawnu(self.command('XDG_DATA_HOME={} {} android android-studio'
+                                                 .format(xdg_data_path, UMAKE)))
+        self.expect_and_no_warn("Choose installation path: {}".format(self.installed_path))
+        self.child.sendline("")
+        self.expect_and_no_warn("\[I Accept.*\]")
+        self.accept_default_and_wait()
+        self.close_and_check_status()
+
     def test_custom_install_path(self):
         """We install android studio in a custom path"""
         # We skip the existing directory prompt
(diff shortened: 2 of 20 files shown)


didrocks:
Hey!

Thanks a lot for your merge request. As you can see I've finally succedeed in implementing SNI on the local server which was needed as Dart now refers multiple servers. This creates an easy fixable conflict with your branch, do you mind rebasing on it?

So, I like your branch in general, thanks for working on that one! Here are just 2 small remarks:
- As you described yourself, I'm not a fan either of bash -c… + strip of the quotation. I'm a little bit afraid of side-effects due to this (especially when extending the command later on). I wonder if self.command() can't rather grow an optional parameter (a dictionary of environment variable), that we pass as the first parameter to run_in_umake_dir for medium test (like run_in_umake_dir <dir> "{env variable: value}" <normal umake command>) and we use that in the script to export envs. For large tests, we just export those (and tearDown() should reset them)
- Maybe you can now hardcode ".local/share/umake" in tests/tools.py and import from it, that will make easier if we change again the default path one day :)

Apart from this, the rest looks good to me! Sorry that you have to rebase, but I was working on the test change for some hours already :)
(oh btw, if you want to run the new medium tests from trunk, do not hesitate to docker pull the image as there are new local certificates).

Thanks again, and keep me posted if I can do anything ;)

oijazsh:
I've rebased, hardcoded ".local/share/umake" in to test/tools.py and implemented a method for setting environment variables for large and medium tests.

The only issue I saw in passing the environment variables as an additional parameter to the command() function is that for medium tests it will be a one-off (which is implied), but for large tests it won't be reset until tearDown() is called. I've used a slightly different approach to ensure that medium tests also only 'unset' their environment once tearDown() is called.

didrocks:
Nice work! Everything looks good to me (just look at my small comment on one part of the diff).
Do you mind rebasing on trunk to tidy up the diff on a more logical pieces? (for instance, squashing f38b386 and 04è76b0, or avoiding the  bash call in aa69514 when is then removed). That will make as well the finale review easier for me :)

And so yeah, our commits telescoped! I don't know which approach do you prefer? I went for a simplier design than what I proposed while avoiding the "bash -c" being explicitely passed for each occurence and mangling with escapes, indeed. I like the approach you implemented as well. I've no strong opinion on this, just tell me which one you prefer and one of the commits need to be adapted :)

You only ran the medium tests on those right? If you want, I can setup a parallel testing jenkins job so that I can point to your branch and have a run with it (it's testing pep8, small and large tests). Those will be ran in both i386 and amd64.
Tell me once you have reorganized your commits (and pick your preferred aproach in the env variable) so that I can run those tests in the CI infra.

Thanks again for your work :)

oijazsh:
I rather like your approach over mine as it is simpler and, at least for our current use case, works well enough. Since there does not seem to be a need for the ability to set environments variables once for a whole test, I guess I can scrap commit a26f5e7 entirely and change aa69514 to be based on your approach.

I'll rebase on trunk shortly.

oijazsh:
I've rebased on the latest master branch and run the pep8, small and medium tests. I went with your approach for setting the environment variables. The large tests still need to be run so it'll be grand if you could setup a jenkins job. :smiley:

didrocks:
Looking excellent as usual!

I did reread the commit this morning and all looks great. Let's run now the large tests and I'll keep you posted :)

didrocks:
Here are the large test results, there are some failures: https://jenkins.qa.ubuntu.com/job/umake-trusty-tests-branch/label=ps-trusty-desktop-amd64-1,type=large/lastCompletedBuild/
(on amd64 for instance).

Part of the issue for 2 of them is the command line split and python pexpect, so I would say not your fault and I need to find something. The issue is that now the install path is longer and so it can't find a match. If you have experience with that, any input is welcome.

The 1st issue sounds more interesting, pexpect is like subprocess.call, so waiting for a binary as first argument, which isn't the case in large test (no wrapping). So you need to do something like this: https://github.com/ubuntu/ubuntu-make/blob/master/tests/large/test_android.py#L406

oijazsh:
Real sorry about the xdg_data_install_path() test failing. I had mostly been running that test throughout my work but forgot about it just once I rebased. I really should have caught that sooner.

That said, after pulling the current HEAD, the android NDK medium test fails from me with:
'''
subprocess.CalledProcessError: Command '['sshpass', '-p', 'user', 'ssh', '-o', 'UserKnownHostsFile=/dev/null', '-o', 'StrictHostKeyChecking=no', '-t', '-q', 'user@10.0.0.65', "/home/omer/src/ubuntu-make/tests/tools/run_in_umake_dir /umake 'sshpass -p user ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -t -q user@10.0.0.65 /home/omer/src/ubuntu-make/tests/tools/run_in_umake_dir /umake 'echo $ANDROID_NDK''"]' returned non-zero exit status 127
'''
The 'sshpass ...... /umake' part is appears twice. This is likely because self.command_as_list() is called twice on the same list twice. Does this also happen on your end?

didrocks:
No worry about the test failing, seems like when fixing the large tests introduced by the previous pull request, I didn't pay attention to the fact that command_as_list() was called twice. So, yeah, fixed and rerun large and medium tests :)

pushed to master now.

oijazsh:
Fixed the large test for xdg_data_path. I pushed up a minor commit but does not seem to have shown up here in this pr at all and still does not seem to be mergeable. I guess I should open a new pull request.

PR #155, opened 2015-10-03
docs: needs bzr and python3-dev for own development workflow

aung00:


didrocks:
Thanks for this contribution and help Ubuntu Make to be better!

All looks good, merging your fix then :+1:

PR #159, opened 2015-10-09
Set the ANDROID_HOME and ANDROID_NDK environment variables

### LINKED ISSUE #158: ANDROID_NDK / ANDROID_HOME variables not set
Is there any specific reason why the ANDROID_NDK / ANDROID_HOME environment variables are not set when installing the Android NDK / SDK respectively? Otherwise they should be set, IMO.

sschuberth:
This fixes #158.

### DIFF
--- a/tests/large/test_android.py
+++ b/tests/large/test_android.py
@@ -403,6 +403,11 @@ class AndroidNDKTests(LargeFrameworkTests):
 
         # we have an installed ndk exec
         self.assert_exec_exists()
+        cmd_list = ["echo $ANDROID_NDK"]
+        if not self.in_container:
+            cmd_list.insert(0, ["bash", "-l", "-c"])
+        self.assertEqual(subprocess.check_output(self.command_as_list(cmd_list)).decode("utf-8").strip(),
+                         self.installed_path)
 
         # launch it, send SIGTERM and check that it exits fine
         self.assertEqual(subprocess.check_call(self.command_as_list([self.exec_path, "gcc"]),
--- a/umake/frameworks/android.py
+++ b/umake/frameworks/android.py
@@ -126,11 +126,12 @@ class AndroidSDK(umake.frameworks.baseinstaller.BaseInstaller):
         return self.category.parse_download_link('id="linux-tools"', line, in_download)
 
     def post_install(self):
-        """Add go necessary env variables"""
+        """Add necessary environment variables"""
+        add_env_to_user(self.name, {"ANDROID_HOME": {"value": self.install_path, "keep": False}})
         # add "platform-tools" to PATH to ensure "adb" can be run once the platform tools are installed via
         # the SDK manager
-        add_env_to_user(self.name, {"PATH": {"value": [os.path.join(self.install_path, "tools"),
-                                                       os.path.join(self.install_path, "platform-tools")]}})
+        add_env_to_user(self.name, {"PATH": {"value": [os.path.join("$ANDROID_HOME", "tools"),
+                                                       os.path.join("$ANDROID_HOME", "platform-tools")]}})
         UI.delayed_display(DisplayMessage(_("You need to restart a shell session for your installation to work")))
 
         """Print wiki page message"""
@@ -162,6 +163,9 @@ class AndroidNDK(umake.frameworks.baseinstaller.BaseInstaller):
         return self.category.parse_download_link('<td>Linux {}'.format(tag_machine), line, in_download)
 
     def post_install(self):
+        """Add necessary environment variables"""
+        add_env_to_user(self.name, {"ANDROID_NDK": {"value": self.install_path, "keep": False}})
+
         """Print wiki page message"""
         UI.display(DisplayMessage("NDK installed in {}. More information on how to use it on {}".format(
                                   self.install_path,


didrocks:
Thanks for the PR!

I'm pretty sure that keep=True by default, so you need to specify it being False to override and not prepend the variable as we discussed.

Also, ensuring that we add those env variables in large (tests/large/test_android.py) and ensuring that the method works in medium tests would be appreciated :)
I can give a hand on that if needed!

sschuberth:
[This](https://github.com/ubuntu/ubuntu-make/blob/master/umake/tools.py#L387) explicitly checks for "keep" being present *and* `true`, so I think we should be fine.

Since I'm now also using the `$ANDROID_HOME` variable in the values added to `PATH`, at least the SDK case should be covered by [this](https://github.com/ubuntu/ubuntu-make/blob/master/tests/large/test_android.py#L366). Do you agree this is sufficient for the SDK?

For the NDK, I can look into adding a test.

sschuberth:
BTW: How about some Travis CI integration to run the tests against PRs?

didrocks:
On tools.py: I think you are wrong in your explanation.
dict.get() is trying to fetch the value, fallbacking to the second parameter (if present) as default if the key is not in dict.
Consequence is that no value -> True is returned.
The second part check that the key is in the environment (as we want to eventually prepend to it).

So, with your current code:
- no ANDROID_HOME set already -> we are fine, ANDROID_HOME= will be set
- if ANDROID_HOME was present -> we will have ANDROID_HOME=new_path:$ANDROID_HOME

Agreed that the PATH check covers it for SDK, so only the NDK test is needed.

Travis CI was enabled at some point (see https://travis-ci.org/ubuntu/ubuntu-make). However, the issue is that their machines are still running precise last time I checked, and so, the requirements for python3 modules are not met (see also: commit 16262dead118a5eb6667f4fdbccdc676dbddc06d).
I tried a couple on months ago, and this was still valid. Mind checking?

sschuberth:
> Consequence is that no value -> True is returned.

Doh, you're right, I've fixed that part.

sschuberth:
> Travis CI was enabled at some point

I know too little about Python pip, virtual-env and the like, but to me it seems as if at least Python 3.4 has been [supported since quite a while](https://github.com/travis-ci/travis-ci/issues/1989), independent of the Ubuntu version they use.

didrocks:
Right, but trusty python-apt is needed (I have some fixes here needed for ubuntu make), and even the build package requirements were not available at the time. I'll have a look if things evolved since I last looked at it.

sschuberth:
Thanks for checking Travis again.

More back on topic:

> Agreed that the PATH check covers it for SDK, so only the NDK test is needed.

I just added the environment variable test to the NDK.

didrocks:
Your test won't work in a medium test environment. To run only this one, you can for instance try: ./runtests tests/medium/test_android.py:AndroidNDKContainer.test_default_android_ndk_install.

This should download the docker container for you and run the tests into it.

What you need is to run a command (that will be automatically wrapper in a ssh, and output will get retrieve). I didn't test yet, but something along the line:
self.assertEqual(subprocess.check_output(self.command("echo $ANDROID_NDK")).decode("utf-8"), self.install_path)
should work

sschuberth:
Thanks for the hints. I've got the test basically working now, but it fails as the environment variable is unset. Do I need to reload the environment somehow to make the variable change effective?

didrocks:
ah, indeed, you can use bash -l to have a subshell which sourced the .profile file first

sschuberth:
Still no luck :-(

didrocks:
You mean in large or medium tests?

command = self.command_as_list(["bash", "-l", "-c", "echo $ANDROID_NDK"])
should work on large tests. (no need for additional inside quoting)

sschuberth:
I mean medium tests, I was running the `runtests` line you quoted above. But I guess this is missing to run the test that is actually setting up the environment?

didrocks:
Ok, I've tried different things to come to the most elegant working solution I can think of for now.

1. Rebase on 27507196f2527698718be718ce8f5835dd2816f8 which had the missing "relog" bits missing from the container (ssh doesn't source .profile in that case)

2. we are going to use a different command between "being in the container" (and so, already in a resourced bash profile) and the current machine for large tests needing the relogging.

Replace your line above simply with:
command = self.command_as_list(["echo $ANDROID_NDK"])
if not self.in_container:
    relogging_command = ["bash", "-l", "-c"]
    relogging_command.extend(command)
    command = relogging_command

That should make it work, keep me posted.

sschuberth:
I've rebased and changed your suggested code slightly to reduce the line count. It works now! :-) Thanks a lot for your help!

didrocks:
Looking good! Just added one small comment, please commit --amend and I'll merge your work ;)

didrocks:
Excellent! Thanks again and merging :)

didrocks:
FYI, the code change compared to my paste results in an issue in large test (https://jenkins.qa.ubuntu.com/job/udtc-trusty-tests/2030/label=ps-trusty-desktop-amd64-1,type=large/testReport/junit/tests.large.test_android/AndroidNDKTests/test_default_android_ndk_install/)
as the argument is [['bash', '-l', '-c'], 'echo $ANDROID_NDK'].

So, I reverted to my proposal which is longer, but more correct ;)

sschuberth:
> So, I reverted to my proposal which is longer, but more correct ;)

Oh, thanks for the fix. But out of curiosity, would you mind explaining to a Python-noob why my solution did not fully work?

didrocks:
Oh sure, no worry!
Basically, when you are doing:
            cmd_list.insert(0, ["bash", "-l", "-c"])

You are inserting to cmd_list, at position 0, one element: ["bash", "-l", "-c"]. This is only a single element, hence the finale cmd_list variable being:
[['bash', '-l', '-c'], 'echo $ANDROID_NDK'], which contains then 2 elements (['bash', '-l', '-c'] and 'echo $ANDROID_NDK').

This is instead of extends which extends one array by another, merging the two, and returning: ['bash', '-l', '-c', 'echo $ANDROID_NDK']

Does it make sense? I should have spot it during my review, sorry!

sschuberth:
Absolutely makes sense. I was looking it up somewhere on StackOverflow, to be honest, but must have missed that the problem description was slightly different than mine. But would simply replacing `cmd_list.insert(0, ["bash", "-l", "-c"])` with `cmd_list = ["bash", "-l", "-c"] + cmd_list` have been the easier and more straight forward fix then?

didrocks:
You are duplicating in memory by creating an extra array (which isn't a real issue here as the list is small), but yeah, resulting list is the same :)

PR #161, opened 2015-10-13
Change default install directory to $XDG_DATA_HOME/umake

oijazsh:
New pull request for #81, since pushing new commits did not seem to make the previous one mergeable again.

### DIFF
--- a/tests/large/__init__.py
+++ b/tests/large/__init__.py
@@ -28,7 +28,7 @@ import stat
 import subprocess
 from time import sleep
 from umake.tools import get_icon_path, get_launcher_path, launcher_exists_and_is_pinned, remove_framework_envs_from_user
-from ..tools import LoggedTestCase, get_path_from_desktop_file, is_in_group
+from ..tools import LoggedTestCase, get_path_from_desktop_file, is_in_group, INSTALL_DIR
 
 
 class LargeFrameworkTests(LoggedTestCase):
@@ -40,7 +40,7 @@ class LargeFrameworkTests(LoggedTestCase):
         self.installed_path = ""
         self.framework_name_for_profile = ""
         self.conf_path = os.path.expanduser("~/.config/umake")
-        self.install_base_path = os.path.expanduser("~/.local/share/umake")
+        self.install_base_path = os.path.expanduser("~/{}".format(INSTALL_DIR))
         self.desktop_filename = ""
         self.child = None
         self.additional_dirs = []
--- a/tests/large/__init__.py
+++ b/tests/large/__init__.py
@@ -40,6 +40,7 @@ class LargeFrameworkTests(LoggedTestCase):
         self.installed_path = ""
         self.framework_name_for_profile = ""
         self.conf_path = os.path.expanduser("~/.config/umake")
+        self.install_base_path = os.path.expanduser("~/.local/share/umake")
         self.desktop_filename = ""
         self.child = None
         self.additional_dirs = []
--- a/tests/large/test_android.py
+++ b/tests/large/test_android.py
@@ -230,6 +230,7 @@ class AndroidStudioTests(LargeFrameworkTests):
         self.expect_and_no_warn("\[I Accept.*\]")
         self.accept_default_and_wait()
         self.close_and_check_status()
+        self.assertFalse(self.launcher_exists_and_is_pinned(self.desktop_filename))
 
     def test_custom_install_path(self):
         """We install android studio in a custom path"""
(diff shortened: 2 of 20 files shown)


didrocks:
Thanks! sorry for the inconvenience of getting up a new PR.

I'm starting the tests at the next available spot. You will have the result by tomorrow morning so that we have a clear view on the current status :)

oijazsh:
I just checked the new test results - the unity3d and pycharm professional test still fail as expected. The problem in the pycharm one is the line split due to the path being too long as you said whereas the problem in the unity3d seems to be far more embarrassing one; I made a typo in commit 6ba69a5, "unity-3d" should be "unity3d". What confuses me is why this was not a problem in the medium test.

I am not sure about how to fix the pycharm issue yet. I'll keep you updated in case I find something.

didrocks:
You intriguated (/scared ;)) me on this medium tests passing, so I had a look.
setUp() in the medium test is:
[…]
super().setUp()
# override with container path
self.installed_path = os.path.join(self.install_base_path, "games", "unity3d")

So, the large wrong path (containing "unity-3d") in the large test is overriden here with the correct path, hence the test passing. So nothing wrong here, phew ;)

Keep me posted on the pycharm issue if you can come with anything elegant (I did search on the subject without any strong lead for now)

oijazsh:
For what it is worth, I can not get the pycharm-professional large test to fail on my system with the same error.

didrocks:
yeah, the issue is that the vertical terminal size depends on the setup. I tried multiple trick to set a very large one on jenkins, but didn't get it right apparently

oijazsh:
I managed to get the pycharm test to fail on my system by using the ```dimensions``` argument to pexpect.spawnu() to set the virtual terminal size. In effect, the test fails with the same captured stdout as jenkins when I use:

```self.child = pexpect.spawnu(self.command('{} ide pycharm-professional'.format(UMAKE)), dimensions=(25, 78))```

The test does not run into this error with a higher number of columns, as in:

```self.child = pexpect.spawnu(self.command('{} ide pycharm-professional'.format(UMAKE)), dimensions=(25, 150))```

didrocks:
Good finding! I confirm it from here: https://github.com/pexpect/ptyprocess/blob/master/ptyprocess/ptyprocess.py#L232.

I think that maybe wrapping pexpect.spawnu() in a helper, which then set dimensions=(25, <large_value>) would be useful, what do you think? At least, we would have a way to get the output tested in a controlled way. I'm opened to any other suggestions :)

oijazsh:
That seems like a good solution to me. I am guessing any reasonably large value for the number of columns should do. I'll try it out in a short while.

I am still not sure if this is the reason the test fails on jenkins however; it seems the test passes on the default values of (24, 80) in ptyprocess. Why does it act as if the columns are 78 when running on jenkins?

didrocks:
Yeah, I don't know why jenkins set a smaller shell, but it seems it's using 78 by default (I did get this when testing a couple of month ago. Of course, it has plugin to change the default size, but that's not the right solution IMHO. The best one is to have tests independent of that client size configuration, so wrapping sounds good!

oijazsh:
Made the wrapper and ran it through the small, medium and pep8 tests.

didrocks:
Excellent! (Maybe add a small comment for the "spawn_process" utility?

I'm starting the large tests now on jenkins

didrocks:
Ok, I think I know why I didn't see it: seems like the dimensions is something relatively new. At least, it wasn't in 14.04 pexpect version: https://jenkins.qa.ubuntu.com/view/All/job/umake-trusty-tests-branch/label=ps-trusty-desktop-amd64-1,type=large/

Let me have a look if I can backport it in the ppa

oijazsh:
Ah, sorry that I did not notice that. According to changelogs, it was added in version 4.0 so it is very recent. Only 2 weeks old.

didrocks:
No worry! I've now packaged the new version and it just finished building in the ppa. I did rerun the tests, results available in ~1h30 at most :)

didrocks:
And here we go! (was a little bit more complex due to python2 support than anticipated)
https://jenkins.qa.ubuntu.com/view/All/job/umake-trusty-tests-branch/

The coverage rate stayed stable as well: https://jenkins.qa.ubuntu.com/view/All/job/umake-trusty-branch-collect/8/label=ps-trusty-desktop-amd64-1/artifact/html-coverage/index.html

Everything's fine for merging on my side, any last word before doing it?

oijazsh:
None, looks good from here.

didrocks:
Perfect! merging then :) I think I'll cut next release next week!

PR #164, opened 2015-10-24
Fix user message

jravetch:


jravetch:
@didrocks Thanks for the comments and agreed that we need the message to be a bit more descriptive.  I've updated the message to reflect that.

didrocks:
Thanks for updating, looking good!

Do you mind changing it in the other frameworks impacted as well?

jravetch:
@didrocks I changed all occurrences of the shell restart message.  I also fixed the wording on how to configure debug logging.

didrocks:
Looking good! Thanks for your work here. Just two small things to fix so that it's mergeable:
- don't change the .po files. Those are refreshed automatically regularly. The issue with you changing the strings right now, is that you tell that the old translations are still valid, which isn't the case as the translation should expect now a parameter. And so, ubuntu make would crash because it's adding an unecessary parameter.
Letting the translation file as they are, the translation is automatically marked as outdated and not used until refreshed.
* Once you remove the .po changes, mind squashing your work in only 2 commits? One for user message and another one for the logging one?

Then, I'll happily merge your change! Thanks again ;)

jravetch:
@didrocks Let me make a new PR.

PR #166, opened 2015-10-27
Fix English wording

jravetch:
Making two commits for shell restart and logging


didrocks:
Looking perfect! Merging now. Thanks a lot for your contribution :)

PR #167, opened 2015-10-30
Use NDK_ROOT instead of ANDROID_NDK to point to the Android NDK

sschuberth:
The official environment variable to use is NDK_ROOT.

### DIFF
--- a/tests/large/test_android.py
+++ b/tests/large/test_android.py
@@ -435,7 +435,7 @@ class AndroidNDKTests(LargeFrameworkTests):
 
         # we have an installed ndk exec
         self.assert_exec_exists()
-        cmd_list = ["echo $ANDROID_NDK"]
+        cmd_list = ["echo $NDK_ROOT"]
         if not self.in_container:
             relogging_command = ["bash", "-l", "-c"]
             relogging_command.extend(cmd_list)
--- a/umake/frameworks/android.py
+++ b/umake/frameworks/android.py
@@ -165,7 +165,7 @@ class AndroidNDK(umake.frameworks.baseinstaller.BaseInstaller):
 
     def post_install(self):
         """Add necessary environment variables"""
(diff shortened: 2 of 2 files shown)


didrocks:
Hey, thanks for this patch!

I wonder however what's the official variables? I saw on the Internet usage of ANDROID-NDK, NDK_ROOT and NDK_HOME. However, going to https://developer.android.com/ndk/guides/setup.html, there is no information on what to set.

I also looked at https://developer.android.com/ndk/guides/ndk-build.html and other pages without finding any official info. Maybe you have a handy link for this?

sschuberth:
I was looking for a link to add as a reference, but I have to admit that I also failed to find a definite answer. To me it seems the NDK itself does not require any variables to be set as it will deduce the location from where the running script is located.

All I can say is, based on my personal experience as an Android developer, most environments that depend on the NDK check the `NDK_ROOT` variable, or maybe the `ANDROID_NDK_ROOT` variable, but not an `ANDROID_NDK` variable.

didrocks:
Ok, I trust you then :)

I'm finishing the medium tests infra change today (to run docker tests as part of our jenkins testsuite) and will merge your change then (probably Monday now).
Thanks again!

sschuberth:
As an indicator I just looked at the StackOverflow search results for these keywords:

http://stackoverflow.com/search?q=[android]+NDK_ROOT
(168 results)

http://stackoverflow.com/search?q=[android]+ANDROID_NDK
(132 results)

http://stackoverflow.com/search?q=[android]+ANDROID_NDK_ROOT
(84 results)

So `ANDROID_NDK` indeed seems to be used more than `ANDROID_NDK_ROOT`, which is not what I would have thought, but still if we only define a single variable, it should be `NDK_ROOT`.

didrocks:
Yeah, the metric difference doesn't seem that drastic. But let's switch to NDK_ROOT based on that.

Issue #104, opened 2015-05-12
Add Rust support


mhalano:
Rust is a new language. The support can be installed using the command:
curl -s https://static.rust-lang.org/rustup.sh | sh -s -- --channel=nightly
And there is available the SolidOak (https://sekao.net/solidoak/) a IDE for Rust.

Tinche:
I'll take this.

didrocks:
@Tinche: thanks for looking into this! Just a word on warning here: I'm varying about running a .sh script rather than unpacking which would be the preferred method (running scripts like that during an installation from things not coming from the archive can violate the ubuntu policy). Do you think we can find a prepared tar that can be just unpack?

Tinche:
There are archives here: https://www.rust-lang.org/install.html, I have this implemented. The problem is, the archives themselves contain an install script, so I'm not sure what exactly to do about this.

As far as I can tell, we actually need to change the global system state (i.e. root access) for the toolchain to work as intended. There are libraries that need to be in a specific place, and ld.conf.d is global-only right?

didrocks:
For the script, if this is the same kind of installer than unity3D (meaning a script including the archive file itself), we have some support in Ubuntu Make to easily deal with that case. Just look at my Unity 3D branch once it's up.

On the toolchain, argh on changing the global system state. Yeah, ld.conf.d is global. Any way we can get things done thanks to a wrapper using LD_LIBRARY_PATH or such?

jravetch:
@didrocks @Tinche  Any update on getting rust working with umake?  Would love to see this working.

Tinche:
I've pushed what I have (which is web page parsing and downloading) to https://github.com/Tinche/ubuntu-make/commit/6fd3b27cea08132576f4a883c62b1d9e44dd6220, if it's any help.

Note that I haven't looked at this since the original comments, so the web page might have changed in the mean time.

jravetch:
@Tinche Cool thanks.  I've made a few changes to support the updated download page and add env variables.  See https://github.com/jravetch/ubuntu-make/commit/0954ce93190749cd3cac69401032c8d030300277 It works with those patches. 

@didrocks Having a weird issue with add_env_to_user, where I need to set two PATH variables, one for rustc, the other for cargo.  However, only cargo is getting set.

didrocks:
@Tinche: thanks for your initial work! It goes a long way ;)

@jravetch: answering you on the PR directly.

jravetch:
[...]

PR #172, opened 2015-11-03
Rust

jravetch:
Thanks to the work of @Tinche Rust can now be added as a framework.  I've updated the download url and few tweaks to get Rust installed properly.  Addresses #104


tests/large/test_rust.py
```diff
@@ -0,0 +1,72 @@
+
```
didrocks: Extra empty line here

tests/large/test_rust.py
```diff
@@ -0,0 +1,72 @@
...
+# -*- coding: utf-8 -*-
+# Copyright (C) 2014 Canonical
```
didrocks: Maybe add your name + update year like 2014-2015?

tests/large/test_rust.py
```diff
@@ -0,0 +1,72 @@
...
+#
+# Authors:
+#  Didier Roche
+#  Tin Tvrtković
```
didrocks: Maybe add your name here (same for other files)

tests/large/test_rust.py
```diff
@@ -0,0 +1,72 @@
...
+            compile_command = ["bash", "-l", "rustc /tmp/hello.rs"]
+
+        self.child = spawn_process(self.command('{} rustc'.format(UMAKE)))
+        self.expect_and_no_warn("Choose installation path: {}".format(self.installed_path))
+        self.child.sendline("")
+        self.expect_and_no_warn("Installation done", timeout=self.TIMEOUT_INSTALL_PROGRESS)
+        self.wait_and_close()
+
+        self.assert_exec_exists()
+        self.assertTrue(self.is_in_path(self.exec_path))
```
didrocks: As you are also adding another directory in cargo/bin, I would also assertTrue that one binary there exists + is in path to ensure both paths were added.

tests/large/test_rust.py
```diff
@@ -0,0 +1,72 @@
...
+
+        # compile a small project
+        output = subprocess.check_output(self.command_as_list(compile_command)).decode()
+
+        if self.in_container:
+            self.assertEqual(output, "hello, world\r\n")
+        else:
+            self.assertEqual(output, "hello, world")
```
didrocks: Are you sure there is no trailing \n here but \r\n in the container?

tests/medium/test_rust.py
```diff
@@ -0,0 +1,34 @@
...
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+"""Tests for rust"""
+
+from . import ContainerTests
+import os
+from ..large import test_rust
+
+
+class RustInContainer(ContainerTests, test_rust.GoTests):
```
didrocks: I think it should inherit from test_ruest.RustTests and not GoTests? (did you run medium tests? I guess not :p)

umake/frameworks/rust.py
```diff
@@ -0,0 +1,144 @@
...
+    def get_metadata_and_check_license(self, result):
+        """Override this so we can use BS and fetch the checksum separately."""
+        logger.debug("Fetched download page, parsing.")
+
+        page = result[self.download_page]
+
+        error_msg = page.error
+        if error_msg:
+            logger.error("An error occurred while downloading {}: {}".format(self.download_page_url, error_msg))
+            UI.return_main_screen()
```
didrocks: Should pass 1 in parameter for exit code here.

umake/frameworks/rust.py
```diff
@@ -0,0 +1,144 @@
...
+        link = (soup.find('div', class_="install")
+                .find('td', class_="inst-type", text="Linux (.tar.gz)")
+                .parent
+                .find(text=self.arch_trans[get_current_arch()])
+                .parent
+                .parent)
+
+        if link is None:
+            logger.error("Can't parse the download URL from the download page.")
+            UI.return_main_screen()
```
didrocks: Same here, pass 1.

umake/frameworks/rust.py
```diff
@@ -0,0 +1,144 @@
...
+        download_url = link.attrs['href']
+        checksum_url = download_url + '.sha256'
+        logger.debug("Found download URL: " + download_url)
+        logger.debug("Downloading checksum first, from " + checksum_url)
+
+        def checksum_downloaded(results):
+            checksum_result = next(iter(results.values()))  # Just get the first.
+            if checksum_result.error:
+                logger.error(checksum_result.error)
+                UI.return_main_screen()
```
didrocks: Same, pass 1

umake/frameworks/rust.py
```diff
@@ -0,0 +1,144 @@
...
+                                                       checksum=Checksum(ChecksumType.sha256, checksum),
+                                                       ignore_encoding=True))
+            self.start_download_and_install()
+
+        DownloadCenter([DownloadItem(checksum_url)], on_done=checksum_downloaded, download=False)
+
+    def post_install(self):
+        """Add rust necessary env variables"""
+        add_env_to_user(self.name, {"PATH": {"value": os.path.join(self.install_path, "rustc", "bin")},
+                                    "PATH": {"value": os.path.join(self.install_path, "cargo", "bin")},
```
didrocks: Here, you are creating a dictionary with two times the same key "PATH". That's the rason why you only see the last one being added, as the other doesn't exist (the key is unique in a map/dictionary).

So, you should rather pass:
{"PATH": {value: "{}:{}".format(os.path.join(self.install_path, "rustc", "bin"), os.path.join(self.install_path, "cargo", "bin"))}, …}

umake/frameworks/rust.py
```diff
@@ -0,0 +1,144 @@
...
+                                    "LD_LIBRARY_PATH": {"value": os.path.join(self.install_path, "rustc", "lib")}})
```
didrocks: Please add in the large tests a way a check that the LD_LIBRARY_PATH has been added.

umake/frameworks/rust.py
```diff
@@ -0,0 +1,144 @@
...
+        UI.delayed_display(DisplayMessage(_("You need to restart your current shell session for your {} installation "
+                                            "to work properly").format(self.name)))
+
+    @property
+    def is_installed(self):
```
didrocks: This method shouldn't be required anymore with latest ubuntu make version.

didrocks:
Hey jravetch! Thanks for this pull requests :)

This mostly looks good, I've added some comments and tried to answer the issue you are seeing with env variable.

On the first commits: I would say, let's try to keep the initial attribution to @Tinche as it's where the credit (and blame of course!) is due. I don't know if it's just github hiding the initial commit and your branched from https://github.com/jravetch/ubuntu-make/commit/0954ce93190749cd3cac69401032c8d030300277 (I didn't branch locally to check), but I think if it's not the case that you should start from this work and then fix user message + add tests (medium and large in separate commits, as you have done).

Also, it seems you didn't run the medium tests, right? (they can only fail for multiple reasons: one being the wrong inheritance, and the second one being that you didn't provide any mock assets and certificates like the other frameworks). Please do run them, in addition to the pep8 and the large tests. If you need help to get them running, do not hesitate to ask (you can only run one category or even one tests. Running the whole testsuite can be long ;)).

You need as well, as the Rust framework is overriding generic functionality of the BaseInstaller, to add more tests to ensure that errors on bad inputs are triggered as expect. Those are of course only mockable in the medium tests (as it's where we have the fake webserver and assets). Basically, in that case I would see:
- a 404 for download page
- an invalid download page content
- a 404 for checksum link
- an invalid checksum content

For the invalid contents, you can look at what is done for instance in tests/medium/test_ide.py:ArduinoIDEInContainer

(yeah, it seems that those 4 tests should be added as well to other test_ide.py frameworks, everytime we derived from BaseInstaller to do something particular, but that's a story for another PR).

Do not hesitate if anything isn't clear enough! I'm happy to give a hand :)
Thanks again!

jravetch:
@didrocks Thanks very much for the comments.  Yeah I didn't branch properly from the original commit from @Tinche.  I'm gonna close this PR to start with a clean branch.

PR #177, opened 2015-11-06
Fix go env

jravetch:
Set keep: "False" so that we don't add a second $GOROOT to the env variable


didrocks:
hum, did you branch from your rust master? look at the diff and number of commits ;)

Please start a new branch for upstream origin/master ;)

didrocks:
(FYI if you git push -f, you don't need to close and reopen a PR)

PR #178, opened 2015-11-06
Override GOROOT value

jravetch:
Set keep: "False" so that we don't end up with a second $GOROOT.


didrocks:
Looking good (and retried locally). Merging! Big thanks :)

Issue #179, opened 2015-11-08
Old version of Unity3D


eldarkg:
[Available a new version](http://forum.unity3d.com/threads/unity-on-linux-release-notes-and-known-issues.350256/#post-2298239)

eldarkg:
Merged

PR #181, opened 2015-11-08
Update unity3d version up to 5.1.0f3+2015091501

eldarkg:
Fix issue #179


didrocks:
Thanks a lot! This looks good and large tests pass.
I'll just modify the fake asset for the medium tests as the name changed. Once the unity team will provide a stable name or download page for the Linux version, we will be able to fetch latest unity version without having to harcode the path (and so, change the medium tests).

Merging itn, should be released before the end of week, thanks again!

eldarkg:
ok

PR #190, opened 2015-11-17
Translate RU

eldarkg:
Update ru translation


didrocks:
Of course, I don't speak russian, but at least, the {} and other parameters are in the correct positions and numbered. I noticed that you refreshed the .po RU template as well, great!

Nice work and thanks a lot, merging ;)

eldarkg:
OK :)

PR #209, opened 2015-12-11
Update Unity3D up to ver.5.2.2f1+20151018

eldarkg:


didrocks:
Thanks a lot! merging it now (and changing medium tests for it).

Do you know if there is any stable page that we can scrap to get latest version of Ubuntu Make (as we do with other frameworks)?

For now, picking this one, thanks again :)

eldarkg:
Hello. I catch links from [here](http://forum.unity3d.com/threads/unity-on-linux-release-notes-and-known-issues.350256).

PR #223, opened 2015-12-23
Update unity3d

LyzardKing:
Downloads the latest unity3d installer. Uses BeautifulSoup to parse the linux support page (http://forum.unity3d.com/threads/unity-on-linux-release-notes-and-known-issues.350256/).
It's not the best solution, but it works for now.


didrocks:
Hey @LyzardKing
Thanks for this patch. This looks good (and thanks for the parsing instead of harcoding the value).

Maybe, we can standardize as we are not harcoding anymore to what is used in other frameworks:
- if you look at the Stencyl framework, in the same file, you can see that download_page is a parameter of the __init__ parameter.
- if you hook it up this way, you can then just remove download_provider_page() override and use    def parse_download_link(self, line, in_download):

basically, this function is called for each line, if you find a match, you return the tuple ((url, None), in_download).
"in_download" value will be returned to the next parse_download_link call (that can enable you to detect multi-line parsing for instance).

Then, you can adapt the medium tests with new assets in tests/data/server-content/ to serve the same and similar files.

If you need any help for any step, do not hesitate! Thanks for this pull request :)

LyzardKing:
The problem is that parse_download_link returns the first installer link in the page. Being a forum post we need the last link to get the latest version of unity3d

didrocks:
Indeed, you're right.

I see two options then:
- I can add a parameter to tell to continue parsing until the end and continue calling parse_download_link even when a match was found (that can be attached to the framework). That way, the last valid content will win.
- you can otherwise override that method:
```
@MainLoop.in_mainloop_thread
```
and grab the checksum (I see there is one in this forum post) + the last download link. Thoughts?

LyzardKing:
I'm not sure. Could a parameter be useful in the future, if other installers present something similar?
For unity3d it's a temporary fix, they should add a standard download page sooner or later... 
The benefit would be also to avoid copying the method to override.
The parameter could be set to False by default.

didrocks:
Sounds good then, I'll implement that shortly. Then, you can adapt to use this code (with this new optional parameter), and add the shasum regexp match.

We'll work on the medium tests together if needed later on :)
I'll keep you posted! Thanks again!

didrocks:
I just pushed https://github.com/ubuntu/ubuntu-make/commit/e69d78c6a7aa33a8b4ba4f6f7ebb71a4a501678a to unblock you. I'll add some tests tomorrow morning for it.

LyzardKing:
Thanks! For now I added the shasum regexp match.

LyzardKing:
It stops working after the download, saying the file has no name. Not sure if it's my connection(very unstable at the moment) or something I got wrong. Checking

LyzardKing:
I forgot to set the checksum_type to ChecksumType.sha1.
It's working now.
For the tests, I changed the file in data/server-content and medium/test_games.py.
Other than that I'm not sure how to deal with the tests.

didrocks:
Your changes looks excellent, no comment to do on them :)

So, on the tests, the large one should work (the large one install unity3d really on your machine and do all the kind of things the user would do).
Remains the medium tests, which is using a docker container to "fake" both the OS packages and assets.

1. What we do here is that we ship a fake web page. I suggest you rename the host to `forum.unity3d.com` instead of `download.unity3d.com` and do the same in `tests/data/server-content/`.

2. In that one, you create `threads/unity-on-linux-release-notes-and-known-issues.350256/index.html` which is a download of the forum page.
In that one, change the last download link to (for instance) http://forum.unity3d.com/unity-editor-installer-fake.sh

3. git mv the fake unity editor to that path (`tests/data/server-content/forum.unity3d.com/unity-editor-installer-fake.sh`). It's better to not have any version mentioned here (as we are not tight anymore to a fixed real url). Finally, take the sha1 of that mock asset and file your download page with it.

`./runtests tests/medium/test_games.py:Unity3DInContainer` should just pass (the first run can be quite slow as it's downloading the container from dockerhub).
Do not hesitate if you need any help with this!

LyzardKing:
something fails with docker. Do I need to setup something?

LyzardKing:
the problem seems to be with spawnu:
in tests/tools/__init__.py", line 218, in spawn_process
    return pexpect.spawn(command, dimensions=(24, 250))
TypeError: __init__() got an unexpected keyword argument 'dimensions'


  File "/usr/lib/python3/dist-packages/pexpect/__init__.py", line 510, in __del__
    if not self.closed:
AttributeError: 'spawn' object has no attribute 'closed'

didrocks:
Ah, indeed, if you are not on xenial or wily, you need to use that ppa for tests dependencies: https://launchpad.net/~ubuntu-desktop/+archive/ubuntu/ubuntu-make-builddeps.

It contains a newer version of pexpect (4.0) than the one in trusty which has that argument.

didrocks:
This additional commit looks good. Just a note: you reference http://forum.unity3d.com/unity-editor-installer-fake.sh when you moved the file under forum.unity3d.com/threads/unity-on-linux-release-notes-and-known-issues.350256/<…>.

Once you move it back to the right place (and have the correct pexpect version installed), your medium test should pass.

LyzardKing:
I fixed the link.
I'm in Wily, but the version of pexpect in the repositories is 3.2.1

LyzardKing:
Ok, I installed pexpect from pip and now the test passes!

didrocks:
Correct on the version, I mixed this as I had it for quite a while :)

Confirming as well that the medium and large tests pass! Excellent work! Merging now.
Thanks again for your excellent work here :)

LyzardKing:
Thank you for the help!

LyzardKing:
The last unity3d update (out now) broke the previous implementation.
Now the sha1sum is in another line (2 after the url)

didrocks:
yeah, that's the issue with them not having a fix download page with strict format. Mind pushing another PR (try to handle both case maybe?) and refresh the medium test page?

LyzardKing:
I'm trying to figure it out. The link is still the last one on the page, as is the shasum.
Maybe we could have it check for the two independently? Or we could ask the people responsible for uploading the files (the user Tak on the forum) to write the posts in the same way every time? (at least for the links)

didrocks:
Yeah, keeping the same format everytime would be a great start :)

I still think indeed that we should support both format for now, just in case mistyping happens!

Issue #207, opened 2015-12-11
Umake should list installable packages


didrocks:
Similar than issue #191, but adding a `umake list` command to list all installable packages on this machine.

I guess the output could be something like:

```
umake list
* android:
  - android-studio (default)
  - android-sdk
  - android-ndk (installed)
* ide:
…
```

Bonus point:
`umake list --all` listing all packages that are supported by Ubuntu Make, being installable or not (because we don't support the release, the architectures and such)

```
umake list
* android:
  - android-studio (default)
  - android-sdk
  - android-ndk (installed)
* ide:
 - awesome-ide (not installable on this machine)
…
```

LyzardKing:
Published in master

Issue #191, opened 2015-11-18
Umake should list installed packages and their location.


jfdm:
`umake` should have an option to list the packages that are already installed, and also where they live.

I've noticed that `umake` doesn't sym-link binaries into `$HOME/.local/bin/`, and one has to link manually, or manually add paths to ones shell rc file. This, is more of an annoyance rather than an hindrance, and noticed in other reports that this is being /taken care of/. Being able to quickly work out what is installed and where it is installed would be beneficial. Especially is none default locations are used.

didrocks:
Hey Jan. That's a very good point indeed!

I am still unsold on the `$HOME/.local/bin/` symlink as this one isn't added to the user's `$PATH`. That's why we add it on a per framework bases to `$PATH` for command line application as needed and provides `.desktop` files that integrates well with Unity launchers and other shells to point to the IDEs.
(this enables us to easily provide multiple flavors of pycharm and other ides like that).

However, as you formulated, having a way for the user to introspect and know what frameworks are installed is a really valuable easy feature to implement (we already have the frameworks and paths available in *~/.config/udtc*). So, I guess, we can get something like:
```
$umake list-installed
android-studio: /home/<user>/…
```

Also, we can have (even if `--help` provides that with additional info)
```
$ umake list
android-ndk
android-studio
pycharm
…
```

I would like to nominate this bug for new contributors first as it's an easy way to dive into the code, and eventually to Google Code-in.
Thanks for the suggestion!

jfdm:
Cheers for the quick response. I've used make to install a few things, and glad it is here; installing recent versions of eclipse on an LTS were/is painful.

TBH I go either way wrt to installation location, and this issue is more about /Developer Experience/ and being able to see what I just installed and where it lives.

Although, this is outside the scope of this issue, I see `umake` as a hybrid combination of [GNU Stow](https://www.gnu.org/software/stow/) and [Homebrew](http://brew.sh/) and /Localised PPAs/ and [CheckInstall](https://help.ubuntu.com/community/CheckInstall). At somepoint it might be nice to see what can be borrowed from these projects to enhance the /Developer Experience/.

LyzardKing:
Published in master

PR #224, opened 2016-01-04
Listing argument

Tankypon:
This PR is here to implement features specified in #191 and #207 issues. At the moment, the argument 'list' is not implemented. By the way, this PR introduces a new function to do the job.

### DIFF
--- a/umake/__init__.py
+++ b/umake/__init__.py
@@ -85,6 +85,15 @@ def set_logging_from_args(args, parser):
         _setup_logging()
 
 
+def should_load_all_frameworks(args):
+    """Set partial or complete framework loading condition based on arg"""
+    for arg in args[1:]:
+        if arg in ["-l", "--list", "--list-installed", "--list-available"]:
+            return True
+
+    return False
+
+
 class _HelpAction(argparse._HelpAction):
 
     def __call__(self, parser, namespace, values, option_string=None):
@@ -117,6 +126,11 @@ def main():
 
     parser.add_argument('-r', '--remove', action="store_true", help=_("Remove specified framework if installed"))
 
+    list_group = parser.add_argument_group("List frameworks").add_mutually_exclusive_group()
+    list_group.add_argument('-l', '--list', action="store_true", help=_("List all frameworks"))
+    list_group.add_argument('--list-installed', action="store_true", help=_("List installed frameworks"))
+    list_group.add_argument('--list-available', action="store_true", help=_("List installable frameworks"))
+
     parser.add_argument('--version', action="store_true", help=_("Print version and exit"))
 
     # set logging ignoring unknown options
@@ -124,8 +138,10 @@ def main():
 
     mainloop = MainLoop()
 
-    # load frameworks and initialize parser
-    load_frameworks()
+    # load frameworks
+    load_frameworks(force_loading=should_load_all_frameworks(sys.argv))
+
+    # initialize parser
     cli.main(parser)
 
     mainloop.run()
--- a/umake/frameworks/__init__.py
+++ b/umake/frameworks/__init__.py
@@ -345,6 +345,58 @@ def load_module(module_abs_name, main_category):
             logger.debug("Attach framework {} to {}".format(framework_name, current_category.name))
 
 
+def list_frameworks():
+    """ Return frameworks and categories description as:
+        [
+            {
+                'category_name':
+                'category_description':
+                'is_installed': BaseCategory.NOT_INSTALLED or
+                                BaseCategory.PARTIALLY_INSTALLED or
+                                BaseCategory.FULLY_INSTALLED
+                'frameworks':
+                    [
+                        {
+                            'framework_name':
+                            'framework_description':
+                            'install_path': None or Path string
+                            'is_installed': True or False
+                            'is_installable': True or False
+                            'is_category_default': True or False
+                            'only_for_removal': True or False
+                        },
+                    ]
+            },
+        ]
+    """
+    categories_dict = list()
+    for category in BaseCategory.categories.values():
+        frameworks_dict = list()
+        for framework in category.frameworks.values():
+            new_fram = {
+                "framework_name": framework.prog_name,
+                "framework_description": framework.description,
+                "install_path": framework.install_path,
+                "is_installed": framework.is_installed,
+                "is_installable": framework.is_installable,
+                "is_category_default": framework.is_category_default,
+                "only_for_removal": framework.only_for_removal
+            }
+
+            frameworks_dict.append(new_fram)
+
+        new_cat = {
+            "category_name": category.prog_name,
+            "category_description": category.description,
+            "is_installed": category.is_installed,
+            "frameworks": frameworks_dict
+        }
+
+        categories_dict.append(new_cat)
+
+    return categories_dict
+
+
 def load_frameworks():
     """Load all modules and assign to correct category"""
     main_category = MainCategory()
--- a/umake/frameworks/__init__.py
+++ b/umake/frameworks/__init__.py
@@ -137,9 +137,9 @@ class BaseCategory():
 
 class BaseFramework(metaclass=abc.ABCMeta):
 
-    def __init__(self, name, description, category, logo_path=None, is_category_default=False, install_path_dir=None,
-                 only_on_archs=None, only_ubuntu_version=None, packages_requirements=None, only_for_removal=False,
-                 expect_license=False, need_root_access=False):
+    def __init__(self, name, description, category, force_loading=False, logo_path=None, is_category_default=False,
+                 install_path_dir=None, only_on_archs=None, only_ubuntu_version=None, packages_requirements=None,
+                 only_for_removal=False, expect_license=False, need_root_access=False):
         self.name = name
         self.description = description
         self.logo_path = None
@@ -197,7 +197,7 @@ class BaseFramework(metaclass=abc.ABCMeta):
             pass
 
         # This requires install_path and will register need_root or not
-        if not self.is_installed and not self.is_installable:
+        if not force_loading and not self.is_installed and not self.is_installable:
             logger.info("Don't register {} as it's not installable on this configuration.".format(name))
             return
 
@@ -326,7 +326,7 @@ def _is_frameworkclass(o):
     return inspect.isclass(o) and issubclass(o, BaseFramework) and not inspect.isabstract(o)
 
 
-def load_module(module_abs_name, main_category):
+def load_module(module_abs_name, main_category, force_loading):
     logger.debug("New framework module: {}".format(module_abs_name))
     if module_abs_name not in sys.modules:
         import_module(module_abs_name)
@@ -341,7 +341,7 @@ def load_module(module_abs_name, main_category):
     if current_category not in BaseCategory.categories.values():
         return
     for framework_name, FrameworkClass in inspect.getmembers(module, _is_frameworkclass):
-        if FrameworkClass(current_category) is not None:
+        if FrameworkClass(current_category, force_loading) is not None:
             logger.debug("Attach framework {} to {}".format(framework_name, current_category.name))
 
 
@@ -397,7 +397,7 @@ def list_frameworks():
     return categories_dict
 
 
-def load_frameworks():
+def load_frameworks(force_loading=False):
     """Load all modules and assign to correct category"""
     main_category = MainCategory()
 
@@ -411,7 +411,7 @@ def load_frameworks():
         local_paths.insert(0, environment_path)
 
     for loader, module_name, ispkg in pkgutil.iter_modules(path=local_paths):
-        load_module(module_name, main_category)
+        load_module(module_name, main_category, force_loading)
     for loader, module_name, ispkg in pkgutil.iter_modules(path=[os.path.dirname(__file__)]):
         module_name = "{}.{}".format(__package__, module_name)
-        load_module(module_name, main_category)
+        load_module(module_name, main_category, force_loading)
--- a/umake/frameworks/__init__.py
+++ b/umake/frameworks/__init__.py
@@ -341,7 +341,7 @@ def load_module(module_abs_name, main_category, force_loading):
     if current_category not in BaseCategory.categories.values():
         return
     for framework_name, FrameworkClass in inspect.getmembers(module, _is_frameworkclass):
-        if FrameworkClass(current_category, force_loading) is not None:
+        if FrameworkClass(category=current_category, force_loading=force_loading) is not None:
             logger.debug("Attach framework {} to {}".format(framework_name, current_category.name))
 
 
--- a/umake/frameworks/android.py
+++ b/umake/frameworks/android.py
@@ -78,9 +78,9 @@ class AndroidCategory(umake.frameworks.BaseCategory):
 
 class AndroidStudio(umake.frameworks.baseinstaller.BaseInstaller):
 
-    def __init__(self, category):
+    def __init__(self, **kwargs):
         super().__init__(name="Android Studio", description=_("Android Studio (default)"), is_category_default=True,
-                         category=category, only_on_archs=_supported_archs, expect_license=True,
+                         only_on_archs=_supported_archs, expect_license=True,
                          packages_requirements=["openjdk-7-jdk | openjdk-8-jdk",
                                                 "libncurses5:i386", "libstdc++6:i386", "zlib1g:i386",
                                                 "jayatana"],
@@ -88,7 +88,7 @@ class AndroidStudio(umake.frameworks.baseinstaller.BaseInstaller):
                          checksum_type=ChecksumType.sha1,
                          dir_to_decompress_in_tarball="android-studio",
                          desktop_filename="android-studio.desktop",
-                         required_files_path=[os.path.join("bin", "studio.sh")])
+                         required_files_path=[os.path.join("bin", "studio.sh")], **kwargs)
 
     def parse_license(self, line, license_txt, in_license):
         """Parse Android Studio download page for license"""
@@ -110,16 +110,16 @@ class AndroidStudio(umake.frameworks.baseinstaller.BaseInstaller):
 
 class AndroidSDK(umake.frameworks.baseinstaller.BaseInstaller):
 
-    def __init__(self, category):
+    def __init__(self, **kwargs):
         super().__init__(name="Android SDK", description=_("Android SDK"),
-                         category=category, only_on_archs=_supported_archs, expect_license=True,
+                         only_on_archs=_supported_archs, expect_license=True,
                          packages_requirements=["openjdk-7-jdk | openjdk-8-jdk",
                                                 "libncurses5:i386", "libstdc++6:i386", "zlib1g:i386",
                                                 "jayatana"],
                          download_page="https://developer.android.com/sdk/index.html",
                          checksum_type=ChecksumType.sha1,
                          dir_to_decompress_in_tarball="android-sdk-linux",
-                         required_files_path=[os.path.join("tools", "android")])
+                         required_files_path=[os.path.join("tools", "android")], **kwargs)
 
     def parse_license(self, line, license_txt, in_license):
         """Parse Android SDK download page for license"""
@@ -147,14 +147,14 @@ class AndroidSDK(umake.frameworks.baseinstaller.BaseInstaller):
 
 class AndroidNDK(umake.frameworks.baseinstaller.BaseInstaller):
 
-    def __init__(self, category):
+    def __init__(self, **kwargs):
         super().__init__(name="Android NDK", description=_("Android NDK"),
-                         category=category, only_on_archs='amd64', expect_license=True,
+                         only_on_archs='amd64', expect_license=True,
                          download_page="https://developer.android.com/ndk/downloads/index.html",
                          checksum_type=ChecksumType.sha1,
                          packages_requirements=['clang'],
                          dir_to_decompress_in_tarball="android-ndk-*",
-                         required_files_path=[os.path.join("ndk-build")])
+                         required_files_path=[os.path.join("ndk-build")], **kwargs)
 
     def parse_license(self, line, license_txt, in_license):
         """Parse Android NDK download page for license"""
@@ -176,6 +176,6 @@ class AndroidNDK(umake.frameworks.baseinstaller.BaseInstaller):
 
 class EclipseADTForRemoval(umake.frameworks.baseinstaller.BaseInstaller):
 
-    def __init__(self, category):
+    def __init__(self, **kwargs):
         super().__init__(name="Eclipse ADT", description="For removal only (not supported upstream anymore)",
-                         download_page=None, category=category, only_on_archs=_supported_archs, only_for_removal=True)
+                         download_page=None, only_on_archs=_supported_archs, only_for_removal=True, **kwargs)
--- a/umake/frameworks/go.py
+++ b/umake/frameworks/go.py
@@ -42,13 +42,14 @@ class GoCategory(umake.frameworks.BaseCategory):
 
 class GoLang(umake.frameworks.baseinstaller.BaseInstaller):
 
-    def __init__(self, category):
+    def __init__(self, **kwargs):
         super().__init__(name="Go Lang", description=_("Google compiler (default)"), is_category_default=True,
-                         category=category, only_on_archs=['i386', 'amd64'],
+                         only_on_archs=['i386', 'amd64'],
                          download_page="https://golang.org/dl/",
-                         checksum_type=ChecksumType.sha256,
+                         checksum_type=ChecksumType.sha1,
                          dir_to_decompress_in_tarball="go",
-                         required_files_path=[os.path.join("bin", "go")])
+                         required_files_path=[os.path.join("bin", "go")],
+                         **kwargs)
 
     def parse_download_link(self, line, in_download):
         """Parse Go download link, expect to find a sha and a url"""
--- a/umake/ui/cli/__init__.py
+++ b/umake/ui/cli/__init__.py
@@ -21,6 +21,7 @@
 
 import argcomplete
 from contextlib import suppress
+from gettext import gettext as _
 import logging
 import os
 from progressbar import ProgressBar, BouncingBar
@@ -28,7 +29,7 @@ import readline
 import sys
 from umake.interactions import InputText, TextWithChoices, LicenseAgreement, DisplayMessage, UnknownProgress
 from umake.ui import UI
-from umake.frameworks import BaseCategory
+from umake.frameworks import BaseCategory, list_frameworks
 from umake.tools import InputError, MainLoop
 from umake.settings import get_version
 
@@ -155,6 +156,68 @@ def mangle_args_for_default_framework(args):
     result_args.extend(pending_args)
     result_args.extend(args_to_append)
     return result_args
+
+
+def get_frameworks_list_output(args):
+    """
+    Get a frameworks list based on the arguments. It returns a string ready to be printed.
+    Multiple forms of the frameworks list can ge given:
+        - List with all frameworks
+        - List with just only installed frameworks
+        - List with just installable frameworks
+    """
+    categories = list_frameworks()
+    print_result = str()
+
+    if args.list or args.list_available:
+        # Sort the categories to prevent a random list at each new program execution
+        for category in sorted(categories, key=lambda cat: cat["category_name"]):
+            if category["category_name"] == "main" and len(category["frameworks"]) == 0:
+                continue
+
+            print_result += "{}: {}".format(category["category_name"], category["category_description"])
+
+            cat_is_installed = str()
+            if category["is_installed"] == BaseCategory.NOT_INSTALLED:
+                cat_is_installed = _("not installed")
+            elif category["is_installed"] == BaseCategory.PARTIALLY_INSTALLED:
+                cat_is_installed = _("partially installed")
+            elif category["is_installed"] == BaseCategory.FULLY_INSTALLED:
+                cat_is_installed = _("fully installed")
+
+            if cat_is_installed:
+                print_result = "{} [{}]".format(print_result, cat_is_installed)
+
+            print_result += "\n"
+
+            # Sort the frameworks to prevent a random list at each new program execution
+            for framework in sorted(category["frameworks"], key=lambda fram: fram["framework_name"]):
+                if args.list_available:
+                    if not framework["is_installable"]:
+                        continue
+
+                print_result += "\t{}: {}".format(framework["framework_name"], framework["framework_description"])
+
+                if not framework["is_installable"]:
+                    print_result = _("{} [not installable on this machine]".format(print_result))
+                elif framework["is_installed"]:
+                    print_result = _("{} [installed]".format(print_result))
+
+                print_result += '\n'
+    elif args.list_installed:
+        # Sort the categories to prevent a random list at each new program execution
+        for category in sorted(categories, key=lambda cat: cat["category_name"]):
+            # Sort the frameworks to prevent a random list at each new program execution
+            for framework in sorted(category["frameworks"], key=lambda fram: fram["framework_name"]):
+                if framework["is_installed"]:
+                    print_result += "{}: {}\n".format(framework["framework_name"],
+                                                      framework["framework_description"])
+                    print_result += "\t{}: {}\n".format(_("path"), framework["install_path"])
+
+        if not print_result:
+            print_result = _("No frameworks are currently installed")
+
+    return print_result
 
 
 def main(parser):
--- a/umake/ui/cli/__init__.py
+++ b/umake/ui/cli/__init__.py
@@ -235,6 +235,10 @@ def main(parser):
         arg_to_parse = mangle_args_for_default_framework(arg_to_parse)
     args = parser.parse_args(arg_to_parse)
 
+    if args.list or args.list_installed or args.list_available:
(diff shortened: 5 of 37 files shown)


umake/ui/cli/__init__.py
```diff
@@ -157,6 +157,60 @@ def mangle_args_for_default_framework(args):
     return result_args
 
 
+def list_frameworks(listing, detailled=True):
```
didrocks: Multiple comments on this:
1. I will move that method in `umake/frameworks/__init__.py` (just above `load_frameworks()` definition) as this is a framework functionality
2. Do not forget to annotate the function explaining the parameters and functions
3. For i18n purpose and multi-possible UI in the future, this function would rather return some json dictionary like:
```
{ 'category_name':
  [
    { 'framework_name':
        {
           'detail': detail_name (always filled, then, the output will filter it if detail isn't chosen)
           'install_path': None or path string
           'installable': True or False
           'only_for_removal': True or False
        }
    },
 ]
}
```

umake/ui/cli/__init__.py
```diff
@@ -157,6 +157,60 @@ def mangle_args_for_default_framework(args):
...
+    if listing == "all" or listing == "available":
```
didrocks: listing is thus not a parameter anymore as we always return everything from this function. However, if you need to use something similar in the future, you can have a look at https://github.com/ubuntu/ubuntu-make/blob/master/umake/tools.py#L51 to see how we define a limited set of choice for a parameter.

umake/ui/cli/__init__.py
```diff
@@ -157,6 +157,60 @@ def mangle_args_for_default_framework(args):
...
+        # Sort the categories to prevent a random list at each new program execution
+        for category in sorted(BaseCategory.categories.values(), key=lambda cat: cat.prog_name):
+            # The category must have at least one framework
+            if len(category.frameworks) == 0:
```
didrocks: I would say we still list categories even having no frameworks. (so an empty category). That will avoid puzzling questions if this situation happened like "why is this category not shown when available in shell completion?"

umake/frameworks/android.py
```diff
@@ -76,12 +76,12 @@ def parse_download_link(self, tag, line, in_download):
 
 class AndroidStudio(umake.frameworks.baseinstaller.BaseInstaller):
 
-    def __init__(self, category):
+    def __init__(self, category, force_loading):
```
didrocks: Instead of passing this parameter (now that we have more than one) in each frameworks, I would suggest to replace the signature and call with with:
```
def __init__(self, **kwargs):
    super().__init__(name=…
                              only_on_archs=…
                              **kwargs)
```
and, so, you remove as well the category=category duplication in each framework at the same time!

Do not forget also to modify the tests frameworks which are under `tests/data/*frameworks/`

umake/ui/cli/__init__.py
```diff
@@ -157,6 +157,60 @@ def mangle_args_for_default_framework(args):
...
+
+            if detailled:
+                print("{}: {}".format(category.prog_name, category.description))
+            else:
+                print("{}:".format(category.prog_name))
+
+            # Sort the frameworks to prevent a random list at each new program execution
+            for framework in sorted(category.frameworks.values(), key=lambda fram: fram.prog_name):
+                # Framework doesn't be removal
+                if framework.only_for_removal:
```
didrocks: So, as shown above, we return frameworks marked as only_for_removal. We will only print it though when the framework is installed and will only (deprecated, only for removal) for instance.

umake/ui/cli/__init__.py
```diff
@@ -157,6 +157,60 @@ def mangle_args_for_default_framework(args):
...
+            for framework in sorted(category.frameworks.values(), key=lambda fram: fram.prog_name):
+                if framework.is_installed:
+                    if detailled:
+                        print("{}: {}".format(framework.prog_name, framework.description))
+                        print("  (path: {})".format(framework.install_path))
+                    else:
+                        print("{} (path: {})".format(framework.prog_name, framework.install_path))
+    else:
+        print("Wrong command")
+        exit(0)
```
didrocks: Even if we won't have the if/else/elif anymore due to that structure change, just a rule: we don't `exit` in random function. Rather, raising an exception that we catch in main program workflow.

In addtion, "Wrong command" should rather be used in logger.Error() to have correct output (and printed in stderr) and in that case, we don't exit 0, but rather 2 ;)

umake/ui/cli/__init__.py
```diff
@@ -157,6 +157,60 @@ def mangle_args_for_default_framework(args):
...
+                if framework.only_for_removal:
+                    continue
+
+                if listing == "available":
+                    if not framework.is_installable:
+                        continue
+
+                if detailled:
+                    # Attribute 'description' of 'framework' contains if the category is default
+                    print("  {}: {} ".format(framework.prog_name, framework.description), end="")
```
didrocks: When you will format the output in the calling functions, some rules for i18n support to translate in multiple languages and clearer code:
* this implies to always preeced messages with _("foo bar baz") (note the _)
* you should never do:
```
a="foo"
b="bar"
print(a + b)
```
Note that it's similar to the form you are using:
```
print("foo", end="")
print("bar", end="")
```
* said differently, you prefer using string concatenation with {} form, which can be translated, like:
```
a="foo"
b="bar"
print(_("{} {}").format(a, b))
```
Of course, you can mix:
```
a="foo"
print(_("bar {}").format(a))
```
* Same, you should use "\n" for line breaks in the previous sentence you are using, you should tend to avoid print() alone (this may be needed in some occasion, like when you want to explicitely break a line between categories).

umake/ui/cli/__init__.py
```diff
@@ -157,6 +157,60 @@ def mangle_args_for_default_framework(args):
...
+
+                if not framework.is_installable:
+                    print("(not installable on this machine)", end="")
+                elif framework.is_installed:
+                    print("(installed)", end="")
+
+                print()
+    elif listing == "installed":
+        # Sort the categories to prevent a random list at each new program execution
+        for category in sorted(BaseCategory.categories.values(), key=lambda cat: cat.prog_name):
```
didrocks: Thanks to the fact that you only return a dictionary here with all results, you won't have to duplicate the logic as you are doing here for categories and frameworks.

didrocks:
I guess that's it for the first pass of review. I tried to be as clear as possible. The base logic is there though, so good work!

Do not hesitate if you have any more questions :)

umake/ui/cli/__init__.py
Tankypon: I begin to take in consideration your suggestion :)

I move the function in umake/frameworks/__init__.py and I had annotated the function, but at the moment the description of the return is false. I began to implement the json dictionnary return, can you tell me if I am on the good way?

umake/ui/cli/__init__.py
Tankypon: I understand your idea of return everything and sort the output after. So I removed the listing and detailled arguments.

umake/ui/cli/__init__.py
Tankypon: This is now done :)

umake/frameworks/android.py
Tankypon: I modified the list argument for each framework, this way is better than the previous for the future if we need to add a new mandatory argument.

It remains to modify all the tests frameworks, I will do it soon ;)

umake/ui/cli/__init__.py
Tankypon: Also done in the new implementation

umake/ui/cli/__init__.py
Tankypon: You are right, this was mainly a prototype function to test it quickly. In the future, I will think about what you said :)

umake/ui/cli/__init__.py
Tankypon: I didn't know that. I think I'm not accustomed to translate my program. The character `_` represents the `gettext` library, is that right? 
Thank you for this suggestion, I will apply it now!

> Same, you should use "\n" for line breaks in the previous sentence you are using, you should tend to avoid print() alone (this may be needed in some occasion, like when you want to explicitely break a line between categories).
I agree with you, I remembered when I wrote this line I was not sure about a print() which print nothing ^^

umake/frameworks/android.py
Tankypon: If I forgot nothing, it's done for the tests frameworks

Tankypon:
I have a question about the description of framework, is it necessary to have in the description `(default)` when a framework is the default for a category. It already exists a keyword argument for that, isn't it? This is `is_category_default`. I think we just need to check this kwarg to add `(default)` in the description if we want to display it. If I say that it's because one of the default framework doesn't have `(default)` in his description, this is `RustLang`

umake/frameworks/android.py
Tankypon: I updated this file too! Now all checks have passed by TravisCI

didrocks:
Good remark on the framework description! Yeah, we should totally remove that from harcoded descriptions of all categories, and then appending that programmatically based on `is_category_default`. I wouldn't add that to kwarg though, but rather have a @property which appends that when description is requested.

I think this will fit better another PR (with some small tests), want to do that one? You can finish that one first I would say, just consider that the (default) is appended for you.

Getting on the code review now :)

tests/small/test_frameworks_loader.py
```diff
@@ -1192,7 +1192,7 @@ class _CustomFramework(umake.frameworks.BaseFramework):
 
         def __init__(self):
             super().__init__(name="Custom", description="Custom uninstallable framework",
-                             category=umake.frameworks.MainCategory())
+                             category=umake.frameworks.MainCategory(), force_loading=False)
```
didrocks: hum, force_loading should be False ion the framework by default. I would prefer that you set that on the framework itself and don't touch those tests.

umake/frameworks/__init__.py
```diff
@@ -340,11 +341,57 @@ def load_module(module_abs_name, main_category):
...
+                new_fram["description"] = framework.description
+            else:
+                new_fram["description"] = framework.description[:framework.description.find("(default)") - 1]
+
+            framework_dict.append(new_fram)
+        new_cat["frameworks"] = framework_dict
+
+        cat_dict.append(new_cat)
+
+    return json.dumps(cat_dict)
```
didrocks: sorry, I was probably unclear about that one :) no need to return a real json object, the list of dict (as a bare object is fine (even better) to me. So you can directly return cat_dict (and remove the import) ;)

umake/frameworks/__init__.py
```diff
@@ -340,11 +341,57 @@ def load_module(module_abs_name, main_category):
...
+        new_cat = dict()
+
+        new_cat["name"] = category.prog_name
+        new_cat["description"] = category.description
+
+        framework_dict = list()
+        for framework in category.frameworks.values():
+            new_fram = dict()
+
+            new_fram["name"] = framework.prog_name
```
didrocks: Another (less repetitive form) would be:
```
new_fram = {
  "name": framework.prog_name,
  "only_for_removal": framework.only_for_removal,
  …
}
```
same for category creation above (and that way, you even don't need `= dict()` ;)

umake/frameworks/__init__.py
```diff
@@ -340,11 +341,57 @@ def load_module(module_abs_name, main_category):
...
+            new_fram["only_for_removal"] = framework.only_for_removal
+            new_fram["is_installable"] = framework.is_installable
+            new_fram["is_category_default"] = framework.is_category_default
+            new_fram["install_path"] = framework.install_path if framework.is_installed else None
+
+            if not framework.is_category_default:
```
didrocks: As discussed and you rightly pointed in your comment, don't care about appending (and thus removing) (default), just take the bare description and consider is being done. Maybe still keep the *is_category_default* parameter in the return object in case we want to use it in the formatting (like in bold in some shells…) in the future

umake/frameworks/__init__.py
```diff
@@ -340,11 +341,57 @@ def load_module(module_abs_name, main_category):
...
+          ]
+        }"""
+    cat_dict = list()
+    for category in BaseCategory.categories.values():
+        new_cat = dict()
+
+        new_cat["name"] = category.prog_name
+        new_cat["description"] = category.description
+
+        framework_dict = list()
```
didrocks: that would be:
"frameworks_dict" as it contains multiple frameworks

umake/frameworks/__init__.py
```diff
@@ -340,11 +341,57 @@ def load_module(module_abs_name, main_category):
...
+                {
+                   'detail': detail_name (always filled, then, the output will filter it if detail isn't chosen)
+                   'install_path': None or path string
+                   'installable': True or False
+                   'only_for_removal': True or False
+                }
+            },
+          ]
+        }"""
+    cat_dict = list()
```
didrocks: same than below, cats_dict (to differentiate one element with the list)

umake/frameworks/go.py
```diff
@@ -42,13 +42,11 @@ def __init__(self):
...
 
-    def __init__(self, category):
+    def __init__(self, **kwargs):
         super().__init__(name="Go Lang", description=_("Google compiler (default)"), is_category_default=True,
-                         category=category, only_on_archs=['i386', 'amd64'],
-                         download_page="https://golang.org/dl/",
-                         checksum_type=ChecksumType.sha1,
-                         dir_to_decompress_in_tarball="go",
-                         required_files_path=[os.path.join("bin", "go")])
+                         only_on_archs=['i386', 'amd64'], download_page="https://golang.org/dl/",
```
didrocks: Please don't add additional diff that's not in context :)
I keep some elements alone in their lines when there are multiple of them for easier readiness (that comments is for all frameworks). Here download_page and checkstum_type for instance.

Please only remove category=category and add **kwargs at the end with minimal formatting (that remark is for all frameworks)

didrocks:
Good progress! So yeah, I like it that way much more :)

See my inline comments (maybe you want to do a git rebase -i to remove the undeed commits after doing this as there will be some reverts?)

So, now that you have that (and once the changes done), you have 3 things to do:
- hook that up in the cli UI with the right option to switch force_loading to true, and then, based on this, show all or only installed ones.
On that note, after our discussion with consistency in the UI, I think those should be like --help options, meaning:
```
umake --list
umake --list-installed (or umake --list --installed), I have no strong opinion :)
```
(I guess the option, for consistency reason should be):
- add some small tests in `tests/small/test_frameworks_loader.py` (just add to existing class and frameworks) for various cases, testing your newly list_frameworks() functions. I will try the different tests case (some with no framework, some with one empty category, some with invalid frameworks, some with uninstallable frameworks, some with installed ones) that are already covered in the different test classes you have in that file, and just add some test_list_frameworks_bla…() which is comparing the `list_frameworks()` output with yours.
- finally, add some large tests once 1. is done to compare some UI output (that would be in `tests/large/test_basics_cli.py`)

As always, I'm here to make clear up anything or to give a hand, nice job until now ;)

Tankypon:
> Good remark on the framework description! Yeah, we should totally remove that from harcoded descriptions of all categories, and then appending that programmatically based on is_category_default. I wouldn't add that to kwarg though, but rather have a @property which appends that when description is requested.

> I think this will fit better another PR (with some small tests), want to do that one? You can finish that one first I would say, just consider that the (default) is appended for you.

I would be very interested to do it, naturally in a later PR. At the moment, I will finish this PR :)

tests/small/test_frameworks_loader.py
Tankypon: I correct it! You are right it's better to have by default `force_loading` to `False`

umake/frameworks/__init__.py
Tankypon: I misunderstand your explanation, I correct the return  type of the function, now it's just a dictionnary

umake/frameworks/__init__.py
Tankypon: I see what you want to say, it's more concise.

umake/frameworks/__init__.py
Tankypon: I remove so the formatting of the description. I will see it in an other PR as I said ;)

umake/frameworks/__init__.py
Tankypon: Well see, I didn't see that one! I also modified the name of the categories dictionary, now it's `categories_dict`. To keep same pattern name in variable.

umake/frameworks/__init__.py
Tankypon: Didn't see this note, well it's done as I said two minutes ago

umake/frameworks/go.py
Tankypon: Sorry for my formatting, I'm not yet accustomed to participate to a project like that. It's entirely my fault, I agree with you I just need to modify the lines of code that concern me! I will undo my modifications ;)

Tankypon:
> See my inline comments (maybe you want to do a git rebase -i to remove the undeed commits after doing this as there will be some reverts?)

I will do it soon, you are right.

> hook that up in the cli UI with the right option to switch force_loading to true, and then, based on this, show all or only installed ones. On that note, after our discussion with consistency in the UI, I think those should be like --help options, meaning

This is the next thing I have to do, I will do it soon ;) If I didn't do it at the moment, it was because I didn't know if I have to implement `umake list` or `umake --list`. It would be better to have `umake list` if therest of the interface was `umake install ...`, etc.
`umake --list` is probably the best way to do it at the moment

> add some small tests in tests/small/test_frameworks_loader.py (just add to existing class and frameworks) for various cases, testing your newly list_frameworks() functions. I will try the different tests case (some with no framework, some with one empty category, some with invalid frameworks, some with uninstallable frameworks, some with installed ones) that are already covered in the different test classes you have in that file, and just add some test_list_frameworks_bla…() which is comparing the list_frameworks() output with yours.
> finally, add some large tests once 1. is done to compare some UI output (that would be in tests/large/test_basics_cli.py)

I will implement tests after the rest

didrocks:
That's perfect! I just wanted to sumarize what was left to do so that we are on the same base :)
No pressure at all, there is nothing hurrying, keep up the good work and ping me if you have any questions or anything new!

umake/__init__.py
```diff
@@ -85,6 +85,17 @@ def set_logging_from_args(args, parser):
         _setup_logging()
 
 
+def set_loading_frameworks(args):
+    """Choose if all frameworks are load ignoring any unknown sys.argv options"""
```
didrocks: I would change the comment to:
"""Set partial or complete framework loading condition based on arg"""

umake/__init__.py
```diff
@@ -85,6 +85,17 @@ def set_logging_from_args(args, parser):
...
+    all_frameworks_loading = False
+    for arg in args[1:]:
+        if arg == "-l" or arg == "--list" or arg == "--list-installed" or arg == "--list-available":
```
didrocks: better to use:
```
if arg in ["-l", "--list", "--list-installed", …]:
```

umake/__init__.py
```diff
@@ -85,6 +85,17 @@ def set_logging_from_args(args, parser):
         _setup_logging()
 
 
+def set_loading_frameworks(args):
```
didrocks: This would be better renamed as "should_load_all_frameworks()", returning a boolean, which is then chained outside to load_frameworks().

This is because:
* you don't really "set loading…" as you are loading here, so the name is wrong
* you don't use sys.argv() but arg itself, don't leak internal implementation in description as it's going to be certainly wrong in the future :)

umake/__init__.py
```diff
@@ -116,6 +127,9 @@ def main():
     parser.add_argument("-v", "--verbose", action="count", default=0, help=_("Increase output verbosity (2 levels)"))
 
     parser.add_argument('-r', '--remove', action="store_true", help=_("Remove specified framework if installed"))
+    parser.add_argument('-l', '--list', action="store_true", help=_("List all frameworks"))
+    parser.add_argument('--list-installed', action="store_true", help=_("List installed frameworks"))
+    parser.add_argument('--list-available', action="store_true", help=_("Remove available frameworks"))
```
didrocks: help description is wrong.

umake/__init__.py
```diff
@@ -124,8 +138,10 @@ def main():
 
     mainloop = MainLoop()
 
-    # load frameworks and initialize parser
-    load_frameworks()
+    # load frameworks
+    set_loading_frameworks(sys.argv)
```
didrocks: hum, I guess you are now loading frameworks twice, as it's called as well later on (loading_frameworks()).

Please ensure this returns a boolean, and load frameworks only once, but do it in the correct place ;)

umake/frameworks/__init__.py
```diff
@@ -340,11 +340,59 @@ def load_module(module_abs_name, main_category):
...
-        if FrameworkClass(current_category) is not None:
+        if FrameworkClass(category=current_category, force_loading=force_loading) is not None:
             logger.debug("Attach framework {} to {}".format(framework_name, current_category.name))
 
 
-def load_frameworks():
+def list_frameworks():
+    """ List all frameworks.
+
+        Return a dictionnary. Format of it is:
```
didrocks: Maybe change this line to:
```
Return frameworks and category description as:
```

umake/frameworks/__init__.py
```diff
@@ -340,11 +340,59 @@ def load_module(module_abs_name, main_category):
...
+        ]
+    """
+    categories_dict = list()
+    for category in BaseCategory.categories.values():
+        frameworks_dict = list()
+        for framework in category.frameworks.values():
+            new_fram = {
+                "framework_name": framework.prog_name,
+                "framework_description": framework.description,
+                "install_path": framework.install_path if framework.is_installed else None,
```
didrocks: what's the default if framework is not installed? If we have framework.install_path returning "" or None, we can just use this and delete the if condition.

umake/ui/cli/__init__.py
```diff
@@ -157,6 +158,41 @@ def mangle_args_for_default_framework(args):
     return result_args
 
 
+def get_frameworks(args):
```
didrocks: I would call it "get_frameworks_list(args)"

or even better:
get_frameworks_list(installed_only=False)

And set installed_only based on args.<vars> outside.

also, please add a description :)

didrocks: Of course, you call this function outside only if there is a list argument.

umake/ui/cli/__init__.py
```diff
@@ -157,6 +158,41 @@ def mangle_args_for_default_framework(args):
...
+            # Sort the frameworks to prevent a random list at each new program execution
+            for framework in sorted(category["frameworks"], key=lambda fram: fram["framework_name"]):
+                if args.list_available:
+                    if not framework["is_installable"]:
+                        continue
+
+                print_result += "\t{}: {} ".format(framework["framework_name"], framework["framework_description"])
+
+                if not framework["is_installable"]:
+                    print_result += "({})".format(_('not installable on this machine'))
```
didrocks: Please use here:
```
print_result = _("{} ({}))".format(print_result, _('not installable on this machine'))
```

This way:
* the string is now translatable
* it's better for i18n, because we can switch between different kind of ordering, depending on the locale (as the first part is internationalized, and not just appended)

So, check the same for the other "print_result +=" ones as well.
I think the only place it's valid to use this is when you append a new element to the list (so after a \n)

didrocks:
And here we go! some changes to be done, but I guess it's giving a good insight and we now have the base structure done! Good work :)

Once those changes are done, we will have the tests as we discussed and also, you could use git rebase -i (see my previous comments) to rebase a clean commit history from master.

Do not hesitate if you have any further question!

umake/ui/cli/__init__.py
Tankypon: I updated as you said, correct me if I forget a `print_result +=`

Tankypon:
Now I have to do the tests and after that I'll do `git rebase -i` :)

umake/ui/cli/__init__.py
```diff
@@ -157,6 +158,65 @@ def mangle_args_for_default_framework(args):
...
 
+def get_frameworks_list(args):
+    """
+    Get a frameworks list based on the arguments.
+    Multiple forms of the frameworks list can ge given:
+        - List with all frameworks
+        - List with just only installed frameworks
+        - List with just installable frameworks
+
+    Args:
```
didrocks: I wouldn't really detail the args here, the function is short enough so that someone can understand it. You can remove the stenza then!

umake/frameworks/__init__.py
```diff
@@ -340,11 +340,61 @@ def load_module(module_abs_name, main_category):
...
 
 
-def load_frameworks():
+def list_frameworks():
+    """ Return frameworks and category description as:
+        [
+            {
+                'category_name':
+                'category_description':
+                'is_installed': 0 (NOT_INSTALLED) or 1 (PARTIALLY_INSTALLED) or 2 (FULLY_INSTALLED)
```
didrocks: The int is an applicaiton detail, it's rather depending on the enum naming itself. So document rather the usage of BaseCategory.NOT_INSTALLED, FULL_INSTALLED, PARTIALLY_INSTALLED.

umake/ui/cli/__init__.py
```diff
@@ -157,6 +158,65 @@ def mangle_args_for_default_framework(args):
...
+    Return:
+        Give the frameworks list as a string
+    """
+    categories = list_frameworks()
+    print_result = str()
+
+    if args.list or args.list_available:
+        # Sort the categories to prevent a random list at each new program execution
+        for category in sorted(categories, key=lambda cat: cat["category_name"]):
+            if category["is_installed"] == 0:
```
didrocks: Use the constants (see previous comment) directly here rather than reharcoding the enum. Imagine we add another new state in the middle, this code would need to adapt. If we use the constants, no need! :)

On a side-note, I don't think that the fact that a category is installed or not or partionnaly is of any value for users. For instance, the "ide" category will ever be fully installed or you will be driven insane ;)

umake/ui/cli/__init__.py
```diff
@@ -157,6 +158,65 @@ def mangle_args_for_default_framework(args):
...
+            # Sort the frameworks to prevent a random list at each new program execution
+            for framework in sorted(category["frameworks"], key=lambda fram: fram["framework_name"]):
+                if args.list_available:
+                    if not framework["is_installable"]:
+                        continue
+
+                print_result += "\t{}: {}".format(framework["framework_name"], framework["framework_description"])
+
+                if not framework["is_installable"]:
+                    print_result = _("{} ({})".format(print_result, _("not installable on this machine")))
```
didrocks: actually, this could be:
`print_result = _("{} (not installable on this machine)".format(print_result))`

and same for below, making sense? (sorry, was tired yesterday when giving this example ;))

umake/ui/cli/__init__.py
```diff
@@ -157,6 +158,65 @@ def mangle_args_for_default_framework(args):
     return result_args
 
 
+def get_frameworks_list(args):
```
didrocks: `get_frameworks_list_output()` would finally make sense here (as we are going to give a string ready to be printed)

didrocks:
Just some more small nitpicks, but it looks really good! Keep me posted on the test side after rebasing ;)

Good work!

umake/__init__.py
```diff
@@ -116,6 +125,9 @@ def main():
     parser.add_argument("-v", "--verbose", action="count", default=0, help=_("Increase output verbosity (2 levels)"))
 
     parser.add_argument('-r', '--remove', action="store_true", help=_("Remove specified framework if installed"))
+    parser.add_argument('-l', '--list', action="store_true", help=_("List all frameworks"))
+    parser.add_argument('--list-installed', action="store_true", help=_("List installed frameworks"))
+    parser.add_argument('--list-available', action="store_true", help=_("List installable frameworks"))
```
didrocks: Oh, just one more thing here.
Those options are uncompatible (you are using one or the other). So you need to create a `add_mutually_exclusive_group()` and add them to it, so that argparse will enable to only use one of them.

Look at runtests, I'm using such an example here.

Tankypon:
I updated the PR to the last comments this time, let me know if I forgot something :)

umake/ui/cli/__init__.py
Tankypon: Normally I corrected them

umake/ui/cli/__init__.py
Tankypon: You are right about the example of "ide" category. If you want I can remove it from the display. maybe it would have be nice to have an other option to have a detailed display versus a compact display. But I think for now, there are already enough arguments. Maybe in a future, with the next major release like that:
`umake list --detailled`
`umake list-available --detailled`
`...`

Tankypon:
I do a `git rebase -i` but now I have some of your commits in my PR, I don't know why.. I'll check it!

Tankypon:
I removed all duplicate commits and I did a `git rebase -i` on top of that.

didrocks:
Ah, I was probably not clear enough on the rebase part :)

When you rebase, you can resquash commits so that they make more sense (no change that is reverted then). After the rebase, I think you should have 4-5 commits only on top of master.

Here is some good rebase documentation if you are not familiar with this: https://git-scm.com/docs/git-rebase

Tankypon:
I did a full rebase on master with my branch! Now I have just 5 commits + 1 for the tests.

LyzardKing:
That's great! I was curious so I tried it out :)
The only thing I'd change is the main category: `main:  (not installed)`
I would hide that, since it serves no real purpose.
What do you think?

Tankypon:
> That's great! I was curious so I tried it out :)

Thank you :)

> The only thing I'd change is the main category: main: (not installed)
 I would hide that, since it serves no real purpose.
 What do you think?

I can remove the extra information if a category is installed or not but as didrocks already told me, it's better to display the category with no frameworks:

> I would say we still list categories even having no frameworks. (so an empty category). That will avoid puzzling questions if this situation happened like "why is this category not shown when available in shell completion?"

But I agree with you it's weird to display that at least for the main category (or a category with no framework).

LyzardKing:
I see the usefulness in being able to see every framework.
The difference is that main is not really a framework.
I think having main actually subtracts from the clarity.
A user would not know what main is and think it's an error.
Great work anyway on this!

On another note:
I would put the "installed", "available",... info in square brackets,
to distringuish it from the "(default)" on the language frameworks
(or we could change those). What do you think?

LyzardKing:
This is what I mean:
rust: Rust language (fully installed)
     rust-lang: The official Rust distribution (installed)
scala: The Scala Programming Language (not installed)
     scala-lang: Scala compiler and interpreter (default)
swift: Swift language (not installed)
     swift-lang: Swift compiler (default)

(rust-lang should have the (default) as well...)

didrocks:
Agreed with @LyzardKing about the main category, this one isn't really one, so better to hide it if empty.

To answer on your quesiton by email (better to centralize the questions here so that we can all help you :)):
- your large tests looks good! You are indeeded exercising the needed code
- however, you will reckon yourself that the granualrity of the large tests isn't extra, and you don't know if you show up empty frameworks, and various cases when the frameworks aren't installed and such. That's why we need a finer granularity and yeah, the small tests for this is needed.
Basically, you send some mock data (false args.argv argument, as a simple object having the correct property), and using the existing frameworks combinations in tests_load_frameworks to ensure you are using the different combinations: For instance, no framework available and --list called, no installable frameworks and --list-installed called and so on (there are at least I guess a good 10 tests to add to try all different combinations)… You just compare the output (the dict) with what you should have from some differents cases.

Those unit tests are really important to quickly spot any regression or unexpected behaviors! Making sense?

Tankypon:
Thanks for your suggestion @LyzardKing I love this one with the bracket, I will do it :)

I agree with both of you, the first iteration of my code (one month ago) removed the main category and others categories with no frameworks but after that I misunderstood one of your comment, and I made a wrong correction.. I will adapt my work with just to remove the main category if there is no frameworks! Thank you (both of you).

Your explanations are really great! You are a good person, I mean you always take the time to explain what you mean and what you think: in the issue, in the PR, etc.. I learned a lot of things with this PR, and I continue to learn a lot of things :)
So it made sense for me like that, I will do the small tests as soon as I can ;) With my studies, I cannot work during the day, so it's normal if it's take time sometimes to made a little change. Hope your understand

Tankypon:
Update:

I did the suggestion from @LyzardKing with the brackets and I corrected the display of the main category if there is no frameworks inside. 
At the moment, the changes are made on my local branch, I'll wait to do the little tests before to make the rebase of my branch and after the push ;)

didrocks:
Thanks for the kind comments! No worry on the timeline, there is really no rush for this! Better to have it well implemented than a first version that needs later on fixes! Take your time, studying is more important :)

Tankypon:
> Basically, you send some mock data (false args.argv argument, as a simple object having the correct property), and using the existing frameworks combinations in tests_load_frameworks to ensure you are using the different combinations: For instance, no framework available and --list called, no installable frameworks and --list-installed called and so on (there are at least I guess a good 10 tests to add to try all different combinations)… You just compare the output (the dict) with what you should have from some differents cases.

I'm again here :) Now I'll end what I started! Can you show me one example, I never used Mock to do a test.

LyzardKing:
Hi! You can have a look at the Mock tests in tests/small/test_frameworks_loader.py
There you can search for Mock(), and you'll see some tests.
Basically you create a stub of the thing you want to test, with the values that you will then verify.

Tankypon:
I already found them in the same file you told me but even with that I don't understand exactly how it works. I will take the time to try that ;)

LyzardKing:
Merged in #497

PR #225, opened 2016-01-05
Eclipse generic installer

LyzardKing:
Modified for sha1.

It works(downloads the correct version.)

The test fails, with:
umake: error: argument category: invalid choice: 'ide' (choose from 'dart', 'go', 'web', 'rust', 'games', 'android')


LyzardKing:
I added an option for checksum with sha512 and added it on the eclipse installer.
Now it should always download the latest eclipse java edition and verify with with checksum sha512.
The one thing is if it should install eclipse-java or eclipse-committers version. (I read that the committers is the new standard version..)

didrocks:
Thanks for this great work! It seems you grok the whole Ubuntu Make structure quite easily (adding new checksum type…)

It was one of the thing I wanted to get off the list (being able to always take latest Eclipse and remove the hardcoded version), that rocks! :)

I have some inline comments coming, but more general questions/remarks:
Do you need help with the test failure? It seems to it that the whole ide category fails to load. This can be explained by the missing imports maybe in a controlled environment.

It seems we want default eclipse to be "eclipse java". Eclipse committers seems to be suited for people who wants to hack on Eclipse itself. We can then structure to add more java-based IDE inheriting from your work.

For medium tests, I guess you want to ship `tests/data/server-content/www.eclipse.org/downloads/index.html` with the similar technic (changing url) than what you did in Unity3D.

Do not hesitate if you need any help!

umake/frameworks/ide.py
```diff
@@ -26,11 +26,10 @@
 from contextlib import suppress
 from gettext import gettext as _
 import grp
-from io import StringIO
```
didrocks: I guess something wrong happened when you rebase on latest master. All the imports (addition or deletion) are a revert compared to when I moved VisualStudioCode the the IDE category.

umake/frameworks/ide.py
```diff
@@ -68,69 +67,66 @@ def __init__(self):
...
+                         description=_("Eclipse Java"),
                          category=category, only_on_archs=['i386', 'amd64'],
-                         download_page=None,
+                         download_page='https://www.eclipse.org/downloads/',
+                         checksum_type=ChecksumType.sha512,
                          dir_to_decompress_in_tarball='eclipse',
                          desktop_filename='eclipse.desktop',
                          required_files_path=["eclipse"],
-                         packages_requirements=['openjdk-7-jdk'])
+                         packages_requirements=['openjdk-8-jdk'])
```
didrocks: Can this work with openjdk-7-jdk? 14.04 doesn't have openjdk-8-jdk and I would like to support it until the next LTS is released at least.

umake/network/download_center.py
```diff
@@ -155,6 +155,8 @@ def _report(block_no, block_size, total_size):
                 actual_checksum = self.md5_for_fd(dest)
             elif checksum_type is ChecksumType.sha256:
                 actual_checksum = self.sha256_for_fd(dest)
+            elif checksum_type is ChecksumType.sha512:
+                actual_checksum = self.sha512_for_fd(dest)
```
didrocks: Mind adding a small tests for this? Look at `tests/small/test_download_center.py`, `TestDownloadCenter` class, `test_download_with_sha256sum` for instance (the files that are tested are in `tests/data/server-content/`.

LyzardKing:
openjdk-7 works fine. I personally use openjdk-8, but I see the problem for ubuntu 14.04. Can we say one or the other?
Something did fail rebasing from master, I'll check that now and look at the tests for eclipse and sha512.

I'm not sure what I need to do for the ide tests failing

didrocks:
We can't say one of another unfortunately (maybe something to add in the future, alternatives handling…)
Once you fix the import and other things, just ping me and I will have a look for the failing tests! Thanks :)

LyzardKing:
The tests run successfully!
I fixed the imports and openjdk version.

didrocks:
Excellent! did you see my other inline comments (still available above) about the additional tests?

Also, for the medium test, I guess you need to rename and move the assets :)

tests/data/server-content/www.eclipse.org/downloads/index.html
```diff
@@ -0,0 +1,896 @@
...
+          </ul>
+          <p>The essential tools for any Java developer, including a Java IDE, a Git client, XML Editor, Mylyn, Maven integration and WindowBuilder...</p>
+        </div>
+        <div class="col-sm-8 download">
+          <div class="col-sm-9 downloadLink-icon"><i class="fa fa-download"></i></div>
+          <div class="col-xs-24 col-sm-15 downloadLink-content">
+            <div class="text-center">
+              <p class="margin-bottom">Linux</p>
+              <ul class="list-inline">
+                                  <li><a class="downloadLink" href="download.php?file=/technology/epp/downloads/release/version/point_release/eclipse-java-linux-gtk.tar.gz" title="Download">32 bit</a></li>
```
LyzardKing: This is the link to the mock eclipse tarfile. Is the path incorrect?

tests/small/test_download_center.py
```diff
@@ -243,6 +244,26 @@ def test_download_with_sha256sum(self):
         self.assertIsNone(result.buffer)
         self.assertIsNone(result.error)
 
+    def test_download_with_sha512sum(self):
```
LyzardKing: This is the test for sha512sum.

tests/data/server-content/www.eclipse.org/downloads/index.html
didrocks: Yeah, that should work, there is a redirect feature in the mock server helping with this, I just need to give a second look to it and run the medium test to ensure it works well :)

tests/small/test_download_center.py
didrocks: excellent! looks good :)

didrocks:
I'll give the whole thing a test run tomorrow and keep you posted, but everything looks good. Nice work! :)

LyzardKing:
Thanks! Let me know

didrocks:
I couldn't resist to give it a try, I have to remove the medium eclipse test for ftp (as we don't use ftp anymore, and it's failing as expected).

However, on the large test is failing:
FAIL: Install eclipse from scratch test case
  File "/home/didrocks/work/ubuntu-make/tests/large/test_ide.py", line 60, in test_default_eclipse_ide_install
    self.assert_exec_exists()
  File "/home/didrocks/work/ubuntu-make/tests/large/__init__.py", line 114, in assert_exec_exists
    self.assertTrue(self.path_exists(self.exec_path))
AssertionError: False is not true

and this is because it seems we download the windows version (I have a windows.exe ;)), so the test does his job as it expect "eclipse" binary inside the folder.Mind having a look at the parser if it matches what you expect?

LyzardKing:
Running "umake ide eclipse" downloads the correct file (tar.gz)
Running "./runtests tests/large/test_ide.py:EclipseIDETests" also works

LyzardKing:
I am missing a few of your latest commits, could that be the problem?

LyzardKing:
On my end everything works even after merging your latest commits

umake/frameworks/ide.py
```diff
@@ -68,69 +68,66 @@ def __init__(self):
...
-            elif arch == 'x86_64':
-                download_url = self.DOWNLOAD_URL_PAT.format(arch='-x86_64',
-                                                            suf='')
-            self.download_requests.append(DownloadItem(download_url, Checksum(ChecksumType.md5, md5)))
-            self.start_download_and_install()
+            self.sha512 = res.buffer.getvalue().decode('utf-8').split()[0]
+        DownloadCenter(urls=[DownloadItem(sha512_url, None)], on_done=done, download=False)
+        # It has to wait for the download and decode of the sha512sum
+        while self.sha512 is None:
+            pass
```
LyzardKing: This while loop is needed to make sure that the process is blocked until the checksum is read.
Otherwise download_sha512 will return with the checksum still None.
I'm sure there is a better way to do this.

didrocks: Indeed, that's part of the issue.

The thing is that you can't give a direct list as you need to parse the page first to get this url result and launch the required download (which happens in another thread), and the while loop defeat that purpose.

So, the more elegant way is a little bit more involved, what I would do:
override (parent implementation in baseinstaller.py):
```
    @MainLoop.in_mainloop_thread
```
In that one, strip the license checking (maybe this method would need to be renamed at some point ;)). result is the whole initial download page in a dictionary (listed by url). This is the code which is calling parse_download_link.
Then, when parse_download_link calls, your
```
 DownloadCenter(urls=[DownloadItem(sha512_url, None)], on_done=self.get_sha_and_start_download, download=False)
```
We know we have the download url already at that point, as the sha512 url is based from it.
So, you just need, in the new get_sha_and_start_download callback (you need to have it running in main thread, so, something around the following):

```
@MainLoop.in_mainloop_thread
def det_sha_and_start_download(self, download_result)
    res = download_result[sha512_url]
    sha512 = res.buffer.getvalue().decode('utf-8').split()[0]
    # you get and store self.download_url
    self.download_requests.append(DownloadItem(self.url, Checksum(self.checksum_type, sha512)))
```

That way, no while loop. You need to add of course empty or invalid conditions logging and error as in the initial get_metadata_and_check_license() method. As you know you have a checksum and no license, the method override can be quite simplified.

Does it make sense? I'm happy to maybe add an additional hook at some point when the download url assets are depending on each others, compared to be just in a fixed set of multiple download pages

didrocks:
On the large tests: `umake ide eclipse` definitively downloads the windows version for me (maybe different files served?) Here is the directory content: http://paste.ubuntu.com/14427986/.
A simple `curl https://www.eclipse.org/downloads/` gives http://paste.ubuntu.com/14428005/

And running your regexp in it matches just two lines:
```
<li><a class="downloadLink" href="download.php?file=/technology/epp/downloads/release/mars/1/eclipse-java-mars-1-win32.zip" title="Download">32 bit</a></li>
<li><a class="downloadLink" href="download.php?file=/technology/epp/downloads/release/mars/1/eclipse-java-mars-1-win32-x86_64.zip" title="Download">64 bit</a></li>
```
So, it seems that the eclipse website offers different download page based on geography. IIRC, that was the &r=1 parameter (for "region"), mind having a look if you find anything?

umake/frameworks/ide.py
```diff
@@ -68,69 +68,66 @@ def __init__(self):
...
+        logger.debug("Downloaded SHA512 is {}".format(self.sha512))
+
+    def parse_download_link(self, line, in_download):
+        """Parse Eclipse download links"""
+        url, self.sha512 = (None, None)
+        if "eclipse-java" in line and self.bits in line:
```
LyzardKing: &r=1 picks the best server for the region.
To be sure of the version we can change the if condition to:
if "eclipse-java" in line and "eclipse-java-ee" not in line and "linux" in line and self.bits in line:
This should now get the correct version.

didrocks: Maybe give it a try with the pastebin I gave you to file up the result, but yeah, at least, we will ensure we download a linux version :)

LyzardKing:
I see what you meant with the windows version being the first. On the page it presents to me the linux version is the first.
The new if statement should fix that.

Using get_metadata_and_licence, where should I put the DownloadCenter line for sha512? It seems to stop after downloading the .sha512, ignoring the archive

didrocks:
You should just put it after you get the url

did you ensure you changed the callback to hook up the on_done to the right method?
(that's the `on_done=self.get_sha_and_start_download` part of my example)

LyzardKing:
with the last commit I set download_page to "https://www.eclipse.org/downloads/?osType=linux"
This should solve the problem you were having with the windows version being downloaded

didrocks:
in_download never returns true here (and so, the whole process hang up): I pasted again the curl output in case we have different results: http://paste.ubuntu.com/14428955/

(I wonder if the fact that I'm on xenial isn't recognized as Linux on the eclipse website)

LyzardKing:
could be that xenial is not recognized, but I forced the linux page to be parsed, so it should make no difference

didrocks:
I can give it a deeper look at it if you want, of course ;)

LyzardKing:
that would be very helpful, since I'm not on xenial (and on wily everything seems to be working).
try with:
curl https://www.eclipse.org/downloads/?osType=linux

if I curl https://www.eclipse.org/downloads/ I get the wrong page as well, but umake works.
Forcing  osType=linux works even in curl

didrocks:
Ok, so pulling from https://github.com/ubuntu/ubuntu-make/tree/eclipse-ide-pagefix should work, do you mind doing this and giving a check on your side?

I have:
- force headers to ensure that eclipse serves to us the Linux version.
- add some additional checks if the web page doesn't provides us what we need to avoid deadlocks
- some slight string change and code reorganization
- add a couple of tests to simulate in medium tests bad download page and bad checksum.

With this, the large tests and the 3 medium tests pass here. I have now the Linux version downloaded :)

Please have a look and tell me if you spot anything!

LyzardKing:
I ran the tests again and it's working here as before. If it works on your end perfect.
The checks and reorganization is good.

didrocks:
Perfect, merging then!

Thanks again for your tremendous work (and diving inside the framework ;)), your work here have been really appreciated.
I think I'll cut a release next Tuesday with your two great improvements.

LyzardKing:
Great! Thanks for your help

PR #228, opened 2016-01-07
parse separately for url and checksum

LyzardKing:
updated unity test

Not working at the moment.
I created a second parser for the checksum.

### DIFF
--- a/tests/data/server-content/forum.unity3d.com/threads/unity-on-linux-release-notes-and-known-issues.350256/index.html
+++ b/tests/data/server-content/forum.unity3d.com/threads/unity-on-linux-release-notes-and-known-issues.350256/index.html
@@ -1369,10 +1369,13 @@ Unsupported Installer for Other 64-bit Distributions:<br />
 <b>Build #2015121801</b><br />
 <br />
 Official Installer for 64-bit Ubuntu Linux:<br />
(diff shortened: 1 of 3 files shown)


LyzardKing:
I can't get the unity test to work. It fails with a timeout on password prompt.
The install succeeds and installs the correct version.

didrocks:
Hey! That looks globally good, however, I have some questions. I can't understand why you override get_metadata_and_check_license. It seems to me that you can quite easily factorize `parse_shasum_link` as well in `parse_download_link` (as you don't need the in_download parameter), or is there anything blocking you?

On the tests with timeout on password prompt: ensure you are using latest docker container that I updated a couple of days ago to not use sshpass (`docker pull didrocks/docker-umake-manual`), and so, working with current ubuntu make trunk.

LyzardKing:
The get_metadata_and_check_license should have been deleted, it was part of a previous test.
Since the link and the shasum are parsed in the same parse_download_link it's not needed.
I'll test it again with the new doker image

LyzardKing:
If I remove the override it will stop after the url is not none.
It doesn't find the shasum and returns an error. I'm not sure why though.
Do you mind seeing if you can find what I'm missing?

LyzardKing:
I added two separate checks for url and checksum not None in get_metadata_and_check_licence in baseInstaller.py.
Now we can see that the parsing does not continue for the checksum after url is read.
Any ideas? or things I got wrong?

didrocks:
does it work as expected with this? (I don't see apart from the message printed) any logical difference?

LyzardKing:
No, it still won't work. The error message I changed was to see what was missing.
It still won't read the checksum

didrocks:
ok, having a look shortly

LyzardKing:
it didn't like the line.endswith. Fixed using in line
Now it works and passes the test.

didrocks:
Confirming that all tests are passing well here as well :)
Thanks for looking at this and your great work again! Merging right now and will be delivered as per next update.

PR #229, opened 2016-01-09
Swift support

LyzardKing:
Added swift framework and tests.
Needs python3-gnupg (to test the signature file).
Depends on clang.

large test working, medium test not working(missing python3-gnupg)


didrocks:
Great! That was something on my list to do. Happy that you tackle it!

I'm giving a first look at the PR. Then, going to propose for you a branch to pull so that medium tests and docker container is prepared (I know this is not trivial to do, I'll explain it at the time). I guess you will then be armed to finish and put everything in shape :)

umake/frameworks/swift.py
```diff
@@ -0,0 +1,120 @@
...
+from gettext import gettext as _
+import logging
+import os
+import re
+import lsb_release
+import gnupg
+
+import umake.frameworks.baseinstaller
+from umake.interactions import DisplayMessage
+from umake.tools import get_current_arch, add_env_to_user, ChecksumType, MainLoop
```
didrocks: get_current_arch and ChecksumType aren't used here, you can remove them

umake/frameworks/swift.py
```diff
@@ -0,0 +1,120 @@
...
+        super().__init__(name="Swift", description=_("Swift language"),
+                         logo_path=None)
+
+
+class SwiftLang(umake.frameworks.baseinstaller.BaseInstaller):
+
+    def __init__(self, category):
+        super().__init__(name="Swift Lang", description=_("Swift compiler (default)"), is_category_default=True,
+                         packages_requirements=["clang", "python3-gnupg"],
+                         category=category, only_on_archs=['amd64'],
```
didrocks: They don't support i386 or other archs?

umake/frameworks/swift.py
```diff
@@ -0,0 +1,120 @@
...
+                         download_page="https://swift.org/download/",
+                         dir_to_decompress_in_tarball="swift*",
+                         required_files_path=[os.path.join("usr", "bin", "swift")])
+        self.release = lsb_release.get_distro_information()['RELEASE']
```
didrocks: tools.get_current_ubuntu_version() achieve something similar, (or we can add it in tools.py if needed)

LyzardKing:
On the download page it says 64 bit only:
"Requirements
Ubuntu 14.04 or 15.10 (64-bit)"

umake/frameworks/swift.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def parse_download_link(self, line, in_download):
+        """Parse Swift download link, expect to find a .sig file"""
+        url_found = False
+        if self.release in line and '.tar.gz.sig' in line:
```
didrocks: I would like this to work on all supported Ubuntu Make release. My approach would be:
for 14.04: take it as expect
for 15.10 and 16.04: take the 15.10 version.

Instead of hardcoding this logic itself (which will quickly get outdated), I guess something like "take previously available version if current one isn't found on page" in this method would rock!

umake/frameworks/swift.py
```diff
@@ -0,0 +1,120 @@
...
+            in_download = True
+        else:
+            in_download = False
+        if in_download:
+            p = re.search(r'href="(.*)" title="PGP Signature"', line)
+            with suppress(AttributeError):
+                self.sig_url = "https://swift.org" + p.group(1)
+                logger.debug(self.sig_url)
```
didrocks: Maybe be a little bit more verbose here: like "Found signature link: {}".format(seg.sig_url)

umake/frameworks/swift.py
```diff
@@ -0,0 +1,120 @@
...
+                url_found = True
+                DownloadCenter(urls=[DownloadItem(self.sig_url, None)],
```
didrocks: In case you do the more advanced logic, it means that you need to move starting the download at line 94 (once the whole file is parsing) under a "if url_found".

umake/frameworks/swift.py
```diff
@@ -0,0 +1,120 @@
...
+            logger.error("An error occurred while downloading {}: {}".format(self.download_page, error_msg))
+            UI.return_main_screen(status_code=1)
+
+        in_download = False
+        url_found = False
+        for line in result[self.download_page].buffer:
+            line_content = line.decode()
+            if not url_found:
+                (_url_found, in_download) = self.parse_download_link(line_content, in_download)
+            if not url_found:
```
didrocks: You have twice `if not url_found:` following each other without assignement in between, so, that can be replaced by with:
```
                (url_found, in_download) = self.parse_download_link(line_content, in_download)
```
(or maybe you can have the logic with "if this version is higher and lower than expected release here…"

umake/frameworks/swift.py
didrocks: Also, some tests (maybe medium assets?) for this logic or shipping that on tools would be interesting?

umake/frameworks/swift.py
```diff
@@ -0,0 +1,120 @@
...
+                url_found = _url_found
+
+        if not url_found:
+            logger.error("Download page changed its syntax or is not parsable")
+            UI.return_main_screen(status_code=1)
+
+    @MainLoop.in_mainloop_thread
```
didrocks: to be testesd, but you shouldn't need that one to run in mainloop

tests/medium/test_swift.py
```diff
@@ -0,0 +1,35 @@
...
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+"""Tests for swift"""
+
+from . import ContainerTests
+import os
+from ..large import test_swift
+
+
+class SwiftInContainer(ContainerTests, test_swift.SwiftTests):
```
didrocks: Looking great, adding some invalid data info tests (like invalid download page, invalid gpg, invalid url) to tests those kinds of errors would be great!

You can have a look at what I added the other day to Eclipse medium tests: https://github.com/ubuntu/ubuntu-make/blob/master/tests/medium/test_ide.py#L45
(Note: we only do that when we override `BaseInstaller` methods, as otherwise, this is covered by `test_baseinstaller.py`)

tests/large/test_swift.py
```diff
@@ -0,0 +1,78 @@
...
+        self.child = spawn_process(self.command('{} swift'.format(UMAKE)))
+        self.expect_and_no_warn("Choose installation path: {}".format(self.installed_path))
+        self.child.sendline("")
+        self.expect_and_no_warn("Installation done", timeout=self.TIMEOUT_INSTALL_PROGRESS)
+        self.wait_and_close()
+
+        self.assert_exec_exists()
+        self.assertTrue(self.is_in_path(self.exec_path))
+        resulting_binary = self.example_prog_dir + "/.build/" + "debug/" + self.example_prog_dir.split('/')[-1]
+        print(resulting_binary)
```
didrocks: debug printing to remove :)

umake/frameworks/swift.py
LyzardKing: previously available would be only for the 'latest' version. Not for the lts release (14.10 would not work).
Is it possible to find the latest lts? or should we hardcode it?

tests/large/test_swift.py
```diff
@@ -0,0 +1,78 @@
...
+        if not self.in_container:
+            self.example_prog_dir = tempfile.mkdtemp()
+            self.additional_dirs.append(self.example_prog_dir)
+            example_file = os.path.join(self.example_prog_dir, "Package.swift")
+            open(example_file, "w").write("")
+            os.mkdir(os.path.join(self.example_prog_dir, "Sources"))
+            example_file = os.path.join(self.example_prog_dir, "Sources", "main.swift")
+            open(example_file, "w").write(self.EXAMPLE_PROJECT)
+            compile_command = ["bash", "-l", "-c", "swift build"]
+        else:  # our mock expects getting that path
```
didrocks: You don't provide a path here contrary to the go one. ;)
I would say "expects getting that command parameter"

didrocks:
Some additional comments:
- tests/data/server-content/swift.org/builds/swift-mock-ubuntu15.10.tar.gz is empty here, was that expected? I guess you need to ship a mock asset with a `usr/bin/swift` binary which will print "Hello world" if the `build` parameter is passed to it (or exit in error otherwise).
- you need to use that binary to sign it and ship the .sign asset as well in the expected path.

I'm going to prepare now the docker image for you.

umake/frameworks/swift.py
```diff
@@ -0,0 +1,118 @@
...
+    def __init__(self):
+        super().__init__(name="Swift", description=_("Swift language"),
+                         logo_path=None)
+
+
+class SwiftLang(umake.frameworks.baseinstaller.BaseInstaller):
+
+    def __init__(self, category):
+        super().__init__(name="Swift Lang", description=_("Swift compiler (default)"), is_category_default=True,
+                         packages_requirements=["clang"],
```
didrocks: From https://swift.org/download, I guess you need to add libicu-dev as well (even if installed by default, in case the user removed it). I'm going to add it to the docker file as well.

umake/frameworks/swift.py
didrocks: I still think that 14.10 (bad case, as not supported) but 15.04 would still work on 14.04 as we don't have the gcc ABI break here. So basically 15.04 -> 14.04 and 16.04 -> 15.10 (until the website publishes a 16.04 version).

What I would do is:
* match all files in the line with the right version syntax and returns this.
* In `get_metadata_and_check_license`, just when calling `self.parse_download_link`, something like:
```
    (new_url_found, in_download) = self.parse_download_link(line_content, in_download)
    if new_url_found > url_found and new_url_found =< <url_link_with_that_version>:
        url_found = new_url_found
```
(you need thus to remove the (if not `url_found` as well). As ubuntu is versioned in a ascii order, you will thus get the closest (but inferior or equals to) the desired version. Making sense?

LyzardKing:
It's not clear to me how to check for 16.04.
I removed the version check from parse_download_link; now it returns every .tar.gz.sig file.
then in get_metadata_and_check_license:

for line in result[self.download_page].buffer:
            line_content = line.decode()
            (new_url_found, in_download) = self.parse_download_link(line_content, in_download)
            if new_url_found and ???
                url_found = new_url_found

didrocks:
the "???" should be something like:
`new_url_found > url_found and new_url_found =< <url_link_with_that_version>:`
as told, the ubuntu versions are ascii ordered, so you should be fine by just compared them as strings. The only thing which is more complex is getting url_link_with_that_version with the correct url.

Another way would be to extract the version from new_url_found, then comparing that with url_found where version is also extracted and current release from the tools function I pointed.

didrocks:
https://github.com/LyzardKing/ubuntu-make/pull/1 should help on getting the medium tests and Travis CI running. I tried to give as many explanations as possible on how things work. I hope that's clear enough :)

LyzardKing:
Can this work for 16.04?
it checks link_version <= running version
It finds 15.10 (good for 15.10 and 16.04)
if not:
it finds 14.04 (good for 14.04)

didrocks:
why do you want to hardcode version? It seems it's a maintainance nightmare for something which is supposed to be really simple by just comparing strings.

I'm happy to propose a PR for this it if you don't see what I mean (once you have done all fixes and merge my first medium PR)

LyzardKing:
I didn't hardcode anything.
I compare self.release with the string from the url

didrocks:
ah ok, seeing your commit now. It's depending on the order links are provided though, we can get something independant, let me comment inline.

LyzardKing:
Thanks for the tests.

didrocks:
it seems you have that figured all out, nice work :)

Do you want me to give another review, or do you want to try large and medium tests first?
I think, once that pass (and the other review done), only the additional corner cases in medium tests will be needed and we will be good to go!

LyzardKing:
I make it import the gpg key from the website.
It would be better to download the .asc file(https://swift.org/keys/all-keys.asc), but I can't get that import to work

LyzardKing:
Just to be sure if you could have another look at it it wouldn't be bad.

tests/large/test_swift.py
```diff
@@ -0,0 +1,77 @@
+
```
didrocks: nitpick: extra empty line here

tests/large/test_swift.py
```diff
@@ -0,0 +1,77 @@
...
+        self.child = spawn_process(self.command('{} swift'.format(UMAKE)))
+        self.expect_and_no_warn("Choose installation path: {}".format(self.installed_path))
+        self.child.sendline("")
+        self.expect_and_no_warn("Installation done", timeout=self.TIMEOUT_INSTALL_PROGRESS)
+        self.wait_and_close()
+
+        self.assert_exec_exists()
+        self.assertTrue(self.is_in_path(self.exec_path))
+        resulting_binary = self.example_prog_dir + "/.build/" + "debug/" + self.example_prog_dir.split('/')[-1]
```
didrocks: os.path.join() ?

umake/frameworks/swift.py
```diff
@@ -0,0 +1,126 @@
...
+class SwiftLang(umake.frameworks.baseinstaller.BaseInstaller):
+
+    def __init__(self, category):
+        super().__init__(name="Swift Lang", description=_("Swift compiler (default)"), is_category_default=True,
+                         packages_requirements=["clang", "libicu-dev"],
+                         category=category, only_on_archs=['amd64'],
+                         download_page="https://swift.org/download/",
+                         dir_to_decompress_in_tarball="swift*",
+                         required_files_path=[os.path.join("usr", "bin", "swift")])
+        self.release = get_current_ubuntu_version()
```
didrocks: get_current_ubuntu_version() is cached, so you can avoid storing it here and replace line 86 with calling the function

umake/frameworks/swift.py
```diff
@@ -0,0 +1,126 @@
...
+            (new_sig_url, in_download) = self.parse_download_link(line_content, in_download)
+            if str(new_sig_url) > str(sig_url):
+                tmp_release = re.search("ubuntu(.....).tar", new_sig_url).group(1)
+                if tmp_release <= self.release:
+                    sig_url = new_sig_url
+
+        if sig_url:
+            DownloadCenter(urls=[DownloadItem(sig_url, None)],
+                           on_done=self.check_gpg_and_start_download, download=False)
+        if not sig_url:
```
didrocks: `else:` instead of `if not sig_url`

didrocks: or even better:
```
if not sig_url:
  logger.error()…
  UI.return…

DownloadCenter(…)
```
To keep coherent with general code style.

LyzardKing:
How do I create the .sig file for the mock archive?

didrocks:
Thanks for fixing as I was reviewing!

So, from what I see there are just the remarks I made that are still visible on https://github.com/ubuntu/ubuntu-make/pull/229/files (check tests/medium/test_swift.py and tests/large/test_swift.py), maybe you missed those 3?
The medium tests will be a little bit more work, but at least, now, you should have the large and medium test passing. Keep me posted!

We can have a look later on (it's not a biggie as long as we don't have any UI that the download is blocking) to get the asc file instead.

didrocks:
oh right, the fact that you import the real .sig is an issue for the archive… A way would be for us to mock the server, but if we are going to get the .asc soon.

Let me look in an hour the python gnupg API if we can import the .asc (and then, we will create the .sig and .asc in the archive)

tests/large/test_swift.py
LyzardKing: swift apparently creates it's own subdirectory, so it's unclear to me how to change it (or if we need this part at all.)

LyzardKing:
Thanks! I'll have a look at the test you made for eclipse then (the invalid one)

tests/large/test_swift.py
didrocks: Sorry, I didn't express that cleanly it seems, I just meant change the comment to:
`# our mock expects getting that command parameter`

which is what your mock asset does (check if there is "build" as parameter and making the proper output)

LyzardKing: Ok. My bad

LyzardKing:
I uploaded the new tests. The problem remaining with the medium test is the missing .sig file on the mock server

didrocks:
yeah, we will add more with an invalid signature once this will be dealt as well.

Also, (in addition to this pep8 issue), it seems you missed my previous comment on the empty line on top of the test file?

LyzardKing:
DownloadCenter(urls=[DownloadItem(sig_url, None)],
                       on_done=self...., download=False)  
if we set download to True is the file saved in /tmp? If so we can add a new DownloadCenter for the .asc
Is there a way then to get the destination path?

didrocks:
yeah, but I think I found a more elegant way, bare with me, working on this as we speak :)

didrocks:
However, even when running directly, I'm getting a gpg issue, did you experience the same?
```
$ LANG=C bin/umake swift
Choose installation path: /home/didrocks/.local/share/umake/swift/swift-lang
ERROR: Unhandled exception
  File "/tmp/ubuntu-make/umake/tools.py", line 158, in wrapper
    function(*args, **kwargs)
  File "/tmp/ubuntu-make/umake/frameworks/swift.py", line 99, in check_gpg_and_start_download
    gpg = gnupg.GPG()
  File "/usr/lib/python3/dist-packages/gnupg.py", line 741, in __init__
    result.stderr))
ValueError: Error invoking gpg: -6: 
gpg: Ohhhh jeeee: ... this is a bug (../../g10/gpg.c:2009:main)
secmem usage: 0/0 bytes in 0/0 blocks of pool 0/65536
```

didrocks:
so, here is my PR: https://github.com/LyzardKing/ubuntu-make/pull/2. This should work in theory, I don't know why the gpg function is getting that error though.

LyzardKing:
I did get that error in the first tests, but that was because I was running the gpg import with root (caused by a typo)

LyzardKing:
I'll try and merge your pr and test it

LyzardKing:
It works. Now the last thing is to recreate the .asc and .sig for the mock server

didrocks:
mind merging it into yours? I need to check if the issue is that it's trying to run it as root here as well (in that case, I need to lock it)

LyzardKing:
What version of gnupg is there in xenial? Wily is a version behind. 
The latest is 0.3.8. 
0.3.7 was released in december 2014...
The error could be solved..

LyzardKing:
Merged

didrocks:
yeah, xenial has 0.3.8 and gnupg is 1.4.19-6ubuntu1. I need to figure that out before we go on the medium tests side (not sure I'll have the time today though)

LyzardKing:
I can't replicate the error any more.. If I can (on a xenial vm?) I'll test a few ideas.

didrocks:
Thanks, I'll still give some look on my side as well. Let's see if we can get something. I ensured that my euid is 1000 (so my user), but still get the same error

didrocks:
also, in ipython3 shell:
```
import gnupg
p = gnupg.GPG()
```
works well. So there is clearly something it doesn't like in ubuntu make environment (and yeah, probably root related, even if euid is fine)

didrocks:
That happens as soon as you call seteuid() before calling gnuepg.GPG(). Simple reproducer:
```
sudo ipython3
import os
import gnupg
os.seteuid(1000) # your user id
gnupg.GPG()
```
-> backtrace. Sounds like it may be due to the module (same issue in 2006 in perl): http://lists.gnupg.org/pipermail/gnupg-users/2006-August/029097.html

didrocks:
I opened https://bitbucket.org/vinay.sajip/python-gnupg/issues?q= for reference

LyzardKing:
Great! I tried to see what fails with euid=1000.
I tried setting in gnupg.GPG() the correct home and binary, but that made no difference.

LyzardKing:
https://lists.gnupg.org/pipermail/gnupg-users/2006-August/029099.html

if we set euid= 1000; then uid != euid.

"This is not a bug, and it certainly isn't new behavior.  GnuPG will
not run if the euid does not match the uid."

didrocks:
I understand their concern, however uid and euid is done for those purpose. I'm afraid though that if we run gnupg as root (without the key being imported), we will end up with some files being root-owned under the user's home.

Do you have time to remove the swift gpg key, for instance, run directly `sudo gpg` to import and verify the key, and check if the files have the expected user's permission, reachable by seahorse and such? If you don't, I can give it a look, but that would be probably tomorrow.

If that works, what I would do as a modification is:
if the needs_root is true for that framework execution (because we are missing some package deps), executes the gpg import part as root, otherwise let it under the current user.

Another way would be to import, executes and remove the swift key.

(for the record this gpg error message is suboptimal in term of debugging though…)

LyzardKing:
Yes, it's not very clear.
I'll have a look at it now.

LyzardKing:
if I run gpg impmort as sudo the ownership is changed to sudo.
Then the user loses access to the file containing the public keys.

LyzardKing:
Large test works on my end!
With and without root

didrocks:
Ah, I'm still having some failures on the large tests (with and without root), it passes the signature now, but swift build is failing:
```
$ ./runtests tests/large/test_swift.py 
[nose.plugins.manager] DEBUG: DefaultPluginManager load plugin nose_json = nose_json.plugin:JsonReportPlugin
[nose.plugins.manager] DEBUG: DefaultPluginManager load plugin cov = nose_cov:Cov
[nose_cov] DEBUG: nose-cov options
[nose_cov] DEBUG: nose-cov configure
bash: swift: command not found
E
======================================================================
ERROR: Install Swift from scratch test case
  File "/home/didrocks/work/ubuntu-make/tests/large/test_swift.py", line 70, in test_default_swift_install
    subprocess.check_call(self.command_as_list(compile_command), cwd=self.example_prog_dir)
  File "/usr/lib/python3.5/subprocess.py", line 581, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['bash', '-l', '-c', 'swift build']' returned non-zero exit status 127

Ran 1 test in 337.351s

FAILED (errors=1)
```

Are you sure the example file is good, maybe your test is passing because you have another swift local installation?

LyzardKing:
I merged your pr with the mock .sig files. For me, the large tests work, the medium test fails.
the executable path should be <install dir>/usr/bin/swift.

didrocks:
hum, are you sure the executable path should be /usr/bin/swift as we install the framework in ~/.local/share/umake/swift/swift-lang by default?

LyzardKing:
not /usr/bin/swift, the path should be ~/.local/share/umake/swift/swift-lang/usr/bin/swift

LyzardKing:
and  ~/.local/share/umake/swift/swift-lang/usr/bin should be added to PATH

didrocks:
ok, this is a recent xenial regression (bash -l doesn't source ~/.profile now). Ignore that one then, it passes on trusty :)

Are you looking at the medium test asset (which is the last issue I guess)?

LyzardKing:
the medium tests run on trusty, right?
If self.exec_path is ~/.local/share/umake/swift/swift-lang/usr/bin/swift
then 
self.assertTrue(self.is_in_path(self.exec_path))
should return true..

didrocks:
yeah, the docker container is running on trusty

LyzardKing:
Is there a way to check the PATH value of the container?
the exec_path is /home/user/.local/share/umake/swift/swift-lang/usr/bin/swift

didrocks:
you comment in tests/medium/__init__.py in tearDown the container removal and then run docker exec (once your retrieve the right container id with docker ps) to connect inside the container. That's the easiest way to debug :)

LyzardKing:
From what I see it fails to change the PATH in the container

didrocks:
I gave it a look, and no, it's not that. You forgot to make the swift binary executable in the mock.
Want me to merge your branch and fix this, or do you want to do it yourself (you will need to resign the mocks then, maybe also, use a symlink from one to the other).

LyzardKing:
Well that was a stupid error on my part...
I'd do it, but I'm not sure how to sign the mocks after.

didrocks:
see the tools I wrote in the PR you merged: https://github.com/LyzardKing/ubuntu-make/commit/01ae6389a518f74dda5ad768f72bc888d2093e8c

Please, make the 386 a symlink to amd64 (I didn't do it, I should have done that). And finally run the medium tests to ensure this works as expected :)

LyzardKing:
what is asc_public_sig in the tool?

didrocks:
(feel free to rephrase the -h if it's not clear): +    asc_public_sig: path to the public gpg key signature.

So, this is the public key exported to the .asc file (which is in swift.org/keys/).
You remove the existing one, run the tool on the first asset, then run the tool on the second asset (the 386 symlink) and commit all this.

LyzardKing:
I created the .sig.
There's one issue still, since yoyu need to run the command "swift build" from the directory.
In the container image, I can't figure where the file is created.
(Swift needs a specific file/folder structure)

didrocks:
Hum, I'm not sure to follow you here. Your mock doesn't manipulate any way file, so you don't need to run/create any.

The mock code you copy is doing:
```
if [ "$*" == "build" ]; then
    echo "Hello, world!"
    exit 0
fi
```

Which is "I expect to have "build" as a parameter, and in that case, I echo the expected output., no file creation expected, nothing?

didrocks:
ah, I think I got what you meant looking at the tests code, you are doing:
```
+        resulting_binary = os.path.join(self.example_prog_dir, ".build", "debug", self.example_prog_dir.split('/')[-1])
 +
 +        subprocess.check_call(self.command_as_list(compile_command), cwd=self.example_prog_dir)
```

So, you need to modify the mock to, instead of echoing "hello world", create a file in a known directory, like /tmp/.build/debug/tmp (don't forget to set it as executable! ;)), and setting first self.example_prog_dir = "/tmp" if running in container.

making sense?

LyzardKing:
Yes, that was my issue. I'll add that now. Thanks

didrocks:
excellent! just push your changes once the medium tests pass, and after a finale review, I'm happy to merge ;)

LyzardKing:
I can't make the directory example_prog_dir on the container image.
if example_prog_dir were "/tmp/hello"
then the swift executable would be:
#!/bin/bash

if [ "$*" == "build" ]; then
    mkdir -p /tmp/hello/.build/debug/hello
    echo -e "echo Hello, world!" > /tmp/hello/.build/debug/hello
    chmod +x /tmp/hello/.build/debug/hello
    exit 0
fi
exit 1

didrocks:
I think the issue is that yo udon't have the shebang placed in `/tmp/.build/debug/tmp` before the echo Hello, world!

You should add the shebang, add a line break (with \n) and then echo
I think then, your executable would work.

LyzardKing:
no, I get the same error even with the fixed echo:

[nose.plugins.manager] DEBUG: DefaultPluginManager load plugin nose_json = nose_json.plugin:JsonReportPlugin
[nose.plugins.manager] DEBUG: DefaultPluginManager load plugin cov = nose_cov:Cov
[nose_cov] DEBUG: nose-cov options
[nose_cov] DEBUG: nose-cov configure
/bin/sh: 1: /tmp/.build/debug/tmp: not found
E.
======================================================================
ERROR: Install Swift from scratch test case
  File "/home/galileo/Documents/Projects/ubuntu-make/tests/large/test_swift.py", line 74, in test_default_swift_install
    output = subprocess.check_output(resulting_binary, cwd=self.example_prog_dir, shell=True).decode()\
  File "/usr/lib/python3.4/subprocess.py", line 620, in check_output
    raise CalledProcessError(retcode, process.args, output=output)
subprocess.CalledProcessError: Command '/tmp/.build/debug/tmp' returned non-zero exit status 127

Ran 2 tests in 23.844s

FAILED (errors=1)

didrocks:
Yeah, but that part was needed :)
I see that you are using `subprocess.check_output()` directly (so, it's trying to executes it on the host and not in the container) where you can see even a few lines above ;) that all commands are wrapped under `self.command()` or `self.command_as_list()`.

This is what wraps all the call in some ssh connection to the container and executing what you expect.
So, you should wrap that way in a self.command(resulting_binary). for the cwd parameter, I would set it only if you are not in the container as well (as the mock doesn't care, and this will have an effect only on host).

LyzardKing:
You're right! with self.command(resulting_binary) it works!
Should I set cwd to None in the container?

LyzardKing:
I ran the pep8/medium/large tests and they work.

didrocks:
Looking good! No need to change cwd if subprocess isn't upset by being called in a directory that doesn't exist on the host :)

Phew, this is now done! Was quite a ride (mostly due to the gpg signature ;)), but glad we have a story for this now!
Thanks again for your excellent work, merging this! I'll announce the next release at UbuCon (LA) next week including your new Eclipse-based IDEs support and this swift one. Nicely done and thanks again! :)

LyzardKing:
Thank you! :)
This has been harder than I had thought.

PR #230, opened 2016-01-12
Eclipse multiple editions

LyzardKing:
Copied from the intellij part. At the moment I added eclipse c/c++ and eclipse php.
Since the icons are all the same I download the ones from the website.
The problem is that they are .png, so not as high quality as the .xpm.
It works, and has tests and mock resources(for eclipse c/c++ and eclipse php).


tests/large/test_ide.py
```diff
@@ -80,6 +80,106 @@ def test_default_eclipse_ide_install(self):
         self.wait_and_close()
 
 
+class EclipsePHPIDETests(LargeFrameworkTests):
```
didrocks: This class is exactly the same than EclipseJavaIDETests with few differences:
installed_path, desktop_filename, command and the "is already installed" (eclipse-php) in umake.

I think we can avoid this duplication by setting up all those parameters in setUp() (you already have two of them) of EclipseJavaIDETests, then inheriting from it and change changing setUp() to match the new php and cpp class, thoughts?

tests/medium/test_ide.py
```diff
@@ -60,6 +60,70 @@ def test_install_with_changed_checksum_page(self):
         self.assertFalse(self.launcher_exists_and_is_pinned(self.desktop_filename))
 
 
+class EclipsePHPIDEInContainer(ContainerTests, test_ide.EclipsePHPIDETests):
```
didrocks: I guess the same can be applied here with inheritance + only changing the desired parameters ;)

umake/frameworks/ide.py
```diff
@@ -139,23 +148,76 @@ def get_sha_and_start_download(self, download_result):
...
-        self.download_requests.append(DownloadItem(url, Checksum(self.checksum_type, sha512)))
+        self.download_requests.append(DownloadItem(url, Checksum(ChecksumType.sha512, sha512)))
         self.start_download_and_install()
 
     def post_install(self):
         """Create the Eclipse launcher"""
-        icon_filename = "icon.xpm"
-        icon_path = join(self.install_path, icon_filename)
+        DownloadCenter(urls=[DownloadItem(self.icon_url, None)],
+                       on_done=self.save_icon, download=True)
```
didrocks: That could be done as the same time that you download the framework itself, line 151, isn't it? If you don't want to override the normal "done" callback for extraction and such, you can just move that line itself under line 151.

Also, that fixes the fact that the launcher icon should be created once the icon is saved at the right place, desktop file refreshement can be racy and won't show up until next reboot.

didrocks:
Excellent work! See my small remarks to avoid code duplication and handling the icon. Other than that, it looks gorgeous!

Thanks for working on it, keep me posted once the changes are done so that I run some tests with this before merging :)

umake/frameworks/ide.py
LyzardKing: I can't get it to download the icon when the path is available (when the archive has been decompressed) if I move the download

didrocks: ah indeed, and it won't be extracted yet so you don't have the destination. Ok, then, let's keep it in post_install for now.

umake/frameworks/ide.py
```diff
@@ -139,23 +148,76 @@ def get_sha_and_start_download(self, download_result):
...
                                                      icon_path=icon_path,
                                                      exec=exec_path,
                                                      comment=comment,
                                                      categories=categories))
 
+    def save_icon(self, download_result):
+        """Save correct Eclipse icon"""
+        icon = download_result.pop(self.icon_url).fd.name
+        shutil.copy(icon, join(self.install_path, self.icon_filename))
+        logger.debug(str(icon))
```
didrocks: A more explicit debug message would be great! :)

tests/large/test_ide.py
LyzardKing: if I move self.command(' ide eclipse-php'.format(UMAKE)) in setUp() I get the following error:
AttributeError: 'EclipsePHPIDEInContainer' object has no attribute 'container_ip'

didrocks: don't do this! Just store the command in setUp(), like:
```
self.command_args = '{} de eclipse-php'.format(UMAKE))
```
and replace `self.command('{} ide eclipse-php'.format(UMAKE)) ` with `self.command(self.command_args)) `

self.command () is executing the command (that's why it's a method and not a variable or overriden property)

LyzardKing: ok, thank you. I missed that completely

tests/large/test_ide.py
```diff
@@ -40,7 +40,9 @@ class EclipseIDETests(LargeFrameworkTests):
     def setUp(self):
         super().setUp()
         self.installed_path = os.path.join(self.install_base_path, "ide", "eclipse")
-        self.desktop_filename = "eclipse.desktop"
+        self.desktop_filename = "eclipse-java.desktop"
+        self.command_args = '{} ide eclipse'.format(UMAKE)
```
didrocks: I see that in medium tests you are calling a similar variable umake_command_args, so maybe just define it here and remove the override in medium tests (as it's specifying the same thing?)

tests/medium/test_ide.py
```diff
@@ -41,25 +41,65 @@ def setUp(self):
...
+                                                        "downloads", "release", "version", "point_release",
+                                                        "eclipse-java-linux-gtk-x86_64.tar.gz.sha512")
 
     def test_install_with_changed_download_page(self):
         """Installing eclipse ide should fail if download page has significantly changed"""
         download_page_file_path = os.path.join(get_data_dir(), "server-content", "www.eclipse.org", "downloads",
                                                "index.html")
-        umake_command = self.command('{} ide eclipse'.format(UMAKE))
-        self.bad_download_page_test(umake_command, download_page_file_path)
+        self.bad_download_page_test(self.command(self.command_args), download_page_file_path)
```
didrocks: that would then become "umake_command_args"

didrocks:
Ah, last thing: I tried as well to run the tests, but it seems you didn't since the refactoring, did you?
```
  File "/tmp/adt-run.TnnEBk/build.GmG/ubuntu-make/tests/medium/test_ide.py", line 59, in test_install_with_changed_checksum_page
    self.bad_download_page_test(self.command(self.command_args), bad_download_page_file_path)
NameError: name 'bad_download_page_file_path' is not defined
```
for instance (alongside other errors).
Please give it a run ;)

LyzardKing:
I think I missed a self. Since the bad_url is now defined in setUp().
I'll test again to check.

didrocks:
I did rerun all the tests and gave another review. Things look good, BUT there is still one test failing.

The issue is simply a typo in one of your last commits, in `tests/large/test_ide.py`, you typed:
`self.test_command = '{} ide eclipse-cpp'.format(UMAKE)` instead of `self.command_args = '{} ide eclipse-cpp'.format(UMAKE)` (and so, it fails on waiting on the "eclipse" path, which is the default).

It's a simple change, I can do it and merge if you want. Otherwise, good work, but as a rule of thumb, always ensure you rerun the tests your impacted when making changes ;)

LyzardKing:
I fixed the typo(sorry) and ran all the tests.

didrocks:
All looks good to me, I started a test run on the infra before clicking on merging (medium and large ide tests)

didrocks:
And tests pass! Thanks a lot for your work here, we now have a stellar Eclipse-based IDEs story ;)

PR #233, opened 2016-01-17
Add manpage that is generated from the --help text

McIntireEvan:
Resolves https://github.com/ubuntu/ubuntu-make/issues/98

Not fully tested yet, but it should work (I am new to this though, so who knows). Either way, I'll fully test it when I wake up.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -7,3 +7,12 @@
 
 override_dh_auto_test:
 	./runtests pep8 small
+
+override_dh_installman:
+	help2man -n "Deploy and setup developers environment easily on ubuntu" -o debian/umake.1 bin/umake
+	sed -i s/+unknown//g debian/umake.1
+	dh_installman
+
+override_dh_clean:
+	rm -f debian/umake.1
+	dh_clean
--- a/debian/control
+++ b/debian/control
@@ -22,6 +22,7 @@ Build-Depends: debhelper (>= 9),
                python3-xdg,
                dbus-x11,
                fakeroot,
+               help2man,
 Maintainer: Didier Roche <didrocks@ubuntu.com>
 Standards-Version: 3.9.6
 X-Python3-Version: >= 3.4
--- a/debian/ubuntu-make.manpages
+++ b/debian/ubuntu-make.manpages
@@ -1,0 +1,1 @@
+debian/umake.1


debian/rules
```diff
@@ -7,3 +7,6 @@
 
 override_dh_auto_test:
 	./runtests pep8 small
+
+override_dh_installman:
+	help2man -o debian/ubuntu-make.1 umake
```
didrocks: I will readd a dh_installman call under this.
```
override_dh_installman:
	help2man -o debian/ubuntu-make.1 umake
        dh_installman
```

That way, the default action is still present.

didrocks:
Hey @McIntireEvan
Nice work as usual! :)

Just some little changes (see above comment) before merging this in. I guess you comitted the generated manpage because it's mandatory to have one file by dh_* for not erroring out (mind checking?), right?

If so, why not shipping an empty file, and ensure that dh_clean as well empty it? We will be sure if something doesn't work one day (like, if we ship an empty man page) that the generation didn't work. Otherwise, we won't notice and we may just ship the older version for some releases which is a bad user experience, making sense?

Thanks for putting this PR up!

debian/rules
McIntireEvan: Thanks, I meant to but totally forgot, nice catch!

didrocks: Ah, this would work only when umake is installed on your system, (but you will print previous version of umake).

However, some things needs tweaking:
- the build fail if ubuntu make isn't installed. Inded, you need to give the path of git umake (bin/umake) and not the system one (otherwise, you would print previous installed verison man page). Check that this path doesn't shows up in the man page (I did, it doesn't seem so) :)
- the man page is called "ubuntu-make" and not the command name "umake". I guess this is simply a flag in help2man
- when running lintian on the built package (with "gbp buildpackage" to build the package from git tree) it prints some lintian warnings:
W: ubuntu-make: manpage-has-useless-whatis-entry usr/share/man/man1/ubuntu-make.1.gz
W: ubuntu-make: manpage-has-errors-from-man usr/share/man/man1/ubuntu-make.1.gz 87: warning [p 2, 1.5i]: cannot adjust line
Maybe those can be fixed with some help2man flags as well?
Also, after a release, the version printed in the man page is <upstream-version>+unknown (this is because we run some system prepared version not in the usual path with modify. I suggested that just after calling help2man, you sed the file to remove any "+unknown".

I guess that's the one I spotted after a quick test. Should be easy to fix them, do not hesitate if you need any help (just a note: I'm travelling tomorrow, and will be in the CEST time for a week)!

McIntireEvan: Wow, I hadn't even considered that, I feel a bit silly! I'll have a new commit after I shower and eat, and hopefully that'll work

McIntireEvan:
Okay, this should be it this time, I wasn't able to find any flags to fix the warnings, maybe with a bit of playing around I could; though I'm in the middle of exam week and I still have my toughest classes to go

If there are any more changes needed please let me know, I'll do them ASAP :) Have fun with your travels!

didrocks:
It looks good, however, you need to give it a build with `gpb buildpackage`. It's failing for me (I guess wrong file target or something like that)
```
help2man -o debian/umake.1 bin/umake
sed -i s/+unknown//g
sed: no input files
debian/rules:12: recipe for target 'override_dh_installman' failed
```

McIntireEvan:
Wow, that's a silly mistake from me - thanks for testing. I'm done with exams finally so I'm going to give it one last go, and everything should be good - thanks again for all the help with this!

McIntireEvan:
Okay, I fixed it all and squashed the commits together. There is still one lintian error from help2man, but I have not been able to find much info on it

didrocks:
This looks awesome and it's working well :)

Getting it into tomorrow's release, merging. Thanks a lot for your work here!

PR #234, opened 2016-01-18
Nodejs support

LyzardKing:
I added nodejs-latest from the nodejs.org website.


tests/large/test_nodejs.py
```diff
@@ -0,0 +1,70 @@
+
```
didrocks: Extra line here, please delete :)

umake/frameworks/nodejs.py
```diff
@@ -0,0 +1,77 @@
...
+            shasum = line.split(' ')[0]
+
+        if url is None and shasum is None:
+            return (None, in_download)
+        return ((url, shasum), in_download)
+
+    def post_install(self):
+        """Add nodejs necessary env variables and move module folder"""
+        add_env_to_user(self.name, {"PATH": {"value": os.path.join(self.install_path, "bin")}})
+        # subprocess.call([os.path.join(self.install_path, "bin", "npm"), "config", "set", "prefix", "~/.node_modules"])
```
didrocks: extra debug to remove I guess :)

didrocks:
That looks great! I have almost no comment to add to tihs :)

In addition to the small cleanup I suggested above (I didn't check for unsued imports yet, as I'm travelling) and I didn't run the tests. I just think that you can enhance them by:
- add the argument check to the "node" mock binary to ensure it is called with "run /tmp/hello.js"
- add another check for the npm binary existence (look at what self.assert_exec_exists() does)
- and finally, add another check for npm called with the right arguments, and that the output is the one we expect.

Apart from that, you added the gpg key, added the needed magic to the docker image (did you rebuild it and tag it locally?). So, everything looks grand from my POV, nice work!

umake/frameworks/nodejs.py
```diff
@@ -0,0 +1,77 @@
...
+        if in_download:
+            url = self.download_page.strip("SHASUMS256.txt") + line.split(' ')[2].rstrip()
+            shasum = line.split(' ')[0]
+
+        if url is None and shasum is None:
+            return (None, in_download)
+        return ((url, shasum), in_download)
+
+    def post_install(self):
+        """Add nodejs necessary env variables and move module folder"""
```
didrocks: It doesn't seem you move module folder anymore, mind fixing the comment?

umake/frameworks/nodejs.py
```diff
@@ -0,0 +1,77 @@
...
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+
+"""Nodejs module"""
+
+from contextlib import suppress
+from gettext import gettext as _
+import logging
+import os
+import subprocess
```
didrocks: subprocess isn't used here

umake/frameworks/nodejs.py
```diff
@@ -0,0 +1,77 @@
...
+import re
```
didrocks: re isn't used as well

umake/frameworks/nodejs.py
LyzardKing: I forgot to remove the comment. That line (75) is needed.
Node creates a folder (node_modules) to store the installed modules.
Normally this folder is created in the same directory as the binary. If we leave it there a reinstall/upgrade with umake would remove the installed modules

umake/frameworks/nodejs.py
LyzardKing: I forgot to uncomment that line.

LyzardKing:
The test for npm should fail, I'm still loking at it.
I did build the image for the tests, but I'm not sure what to do with that at the moment

LyzardKing:
I fixed the npm binary and now it works on the local image.

didrocks:
Nice!

For the medium test to use your image, you just need to tag it.

You can run:
`docker images`
-> note the image id you just created.
`docker tag -f <image_id> didrocks/docker-umake-manual`

Then, you should be able to run the medium tests.

LyzardKing:
ah ok. I changed the name of the image in medium/__init__.py on my system to run the test (I didn't upload that)

didrocks:
that's fine as well :)

Tell me once I should give this another look, I'm at UbuCon, but I'll try to get that in quickly :)

LyzardKing:
On my image the medium test passes.

LyzardKing:
Enjoy UbuCon! If you happen to have some time to look at the pr let me know, but there's no rush :)

didrocks:
This is looking great! I have no further comment on the code.
I'm at a conference and so, the wifi isn't great, but I did try to run the large tests, and that happened:
```
subprocess.CalledProcessError: Command '['bash', '-l', '/home/didrocks/.local/share/umake/nodejs/nodejs-lang/bin/npm', 'config set prefix ~/.node_modules']' returned non-zero exit status 1
```

If you don't reproduce it, I'll wait to be on a better network, but do you mind giving a check?
Thanks a lot!

LyzardKing:
I have a very spotty internet connection now. I'll check later today
or at latest tomorrow.
I think the issue may be with expanding the "~". Could we swap that
with "$HOME"?

didrocks:
So, I rebuilt the docker container and pushed it to the hub, the medium test pass.

What only remain to fix then is this large test (still getting the same error). I don't think it's the expansion TBH as looking at the error, it has /home/didrocks (and we are using os.path.expanduser).

LyzardKing:
I fixed it. It was trying to reset the path for the node modules. Now it checks to see if the set command (in the framework) has succeeded. It works on my machine. Let me know.

didrocks:
Excellent! Working well, everything's good now to merge.

Thanks again for your excellent work on this!

PR #237, opened 2016-01-25
Add python-gnupg to requirements.txt

oijazsh:
Had to add python-gnupg to run the latest master under a virtualenv.


didrocks:
Nicely spotted! Thanks :)

PR #238, opened 2016-01-27
Symlink for editor binaries

LyzardKing:
link to binaries in ~/.bin (still not sure about that, I think ~/.local/share/umake/bin is better)
Implemented for VSCode and eclipse. Needs some better testing before it can be merged.
I added a method in tools.py to add the link and a property to be able to delete the link.


tests/large/test_ide.py
```diff
@@ -60,6 +60,7 @@ def test_default_eclipse_ide_install(self):
         self.assertTrue(self.launcher_exists_and_is_pinned(self.desktop_filename))
         self.assert_exec_exists()
         self.assert_icon_exists()
+        self.assertTrue(self.is_in_path(os.path.join(self.binary_dir, self.desktop_filename.split('.')[0])))
```
didrocks: maybe we can make that an helper in tests/large/__init__.py?

umake/frameworks/baseinstaller.py
```diff
@@ -117,6 +118,9 @@ def remove(self):
         if self.icon_filename:
             with suppress(FileNotFoundError):
                 os.remove(get_icon_path(self.icon_filename))
+        if self.add_link is True:
+            with suppress(FileNotFoundError):
+                os.remove(os.path.join(os.path.expanduser("~"), '.bin', self.desktop_filename.split('.')[0]))
```
didrocks: maybe set the path in settings.py rather than copying it here and when creating it? (we can reuse it as well in the tests that way)

umake/frameworks/ide.py
```diff
@@ -673,7 +675,8 @@ def __init__(self, category):
                          desktop_filename="visual-studio-code.desktop",
                          required_files_path=["Code"],
                          dir_to_decompress_in_tarball="VSCode-linux-*",
-                         packages_requirements=["libgtk2.0-0"])
+                         packages_requirements=["libgtk2.0-0"],
+                         add_link=True)
```
didrocks: I would really just condition the add_link to the existence of desktop_filename and that's it. Same when removing, if we have a desktop_filename, we remove if exists.

umake/tools.py
```diff
@@ -337,6 +337,15 @@ def create_launcher(desktop_filename, content):
         gsettings.set_strv("favorites", launcher_list)
 
 
+def add_exec_link(exec_path, destination_name):
+        bin_folder = os.path.join(os.path.expanduser("~"), '.bin')
```
didrocks: so, expose that one in a property (to reuse the dir when removing). I agree, maybe change for the umake/bin subdirectory rather in local/share.

umake/tools.py
```diff
@@ -337,6 +337,15 @@ def create_launcher(desktop_filename, content):
         gsettings.set_strv("favorites", launcher_list)
 
 
+def add_exec_link(exec_path, destination_name):
```
didrocks: so, this will need small tests (I'll put the removal of links as another utility in tools.py as well), and add of course some inline documentation

didrocks:
Thanks for trying this! Here are some initial thoughts, I mostly agree with the design and I guess it's a good base for the rest!

Good work, keep me posted whenever you have time to change this :)

LyzardKing:
The only thing missing now (if you agree on how this works now) is the small test for the add_exec_link

umake/tools.py
```diff
@@ -337,6 +337,15 @@ def create_launcher(desktop_filename, content):
...
+    bin_folder = settings.DEFAULT_BINARY_LINK_PATH
+    with suppress(FileExistsError):
+        os.mkdir(bin_folder)
```
didrocks: Instead of this, you can use:
os.makedirs(exist_ok=True)

umake/frameworks/ide.py
```diff
@@ -749,9 +750,11 @@ def check_external_license(self, result):
...
         """Create the Visual Studio Code launcher"""
+        exec_path = os.path.join(self.install_path, "Code")
         create_launcher(self.desktop_filename, get_application_desktop_file(name=_("Visual Studio Code"),
                         icon_path=os.path.join(self.install_path, "resources", "app", "resources", "linux",
                                                "vscode.png"),
-                        exec=os.path.join(self.install_path, "Code"),
+                        exec=exec_path,
                         comment=_("Visual Studio focused on modern web and cloud"),
                         categories="Development;IDE;"))
+        add_exec_link(exec_path, self.desktop_filename.split('.')[0])
```
didrocks: I would rather add the add_exec_link call to BaseInstaller, on condition there is a desktop file, and matching its exec path (that can be a new flag in __init__(). What do you think?

didrocks:
Excellent work!

So yeah, the unit tests is one of the last step. However, as written above, I wonder if we couldn't use something like a more global mechanism for this? I guess if there is a desktop file that's going to be created, we want to have a symlink (based on self.exec_path?). Maybe we can write something around this and ensure that everytime you ship into this configuration, the baseinstaller framework is hooking this in. What do you think?

umake/tools.py
LyzardKing: Fixed

LyzardKing:
I agree. I'll have a look at that. I am trying to make it as generic as possible, so that's better.

didrocks:
You rock! :-)

LyzardKing:
Thanks :)
I changed it and it seems to work. The one thing is that the exec_path should now always point to the executable, without options:
Not        exec_path = '"{}" %f'.format(join(self.install_path, "bin", self.executable))
but        exec_path = 'join(self.install_path, "bin", self.executable)

and then in the create_desktop we pass "{}" %f'.format(exec_path).

Could that be good? Needs changing on many frameworks, but it would solve the issue

didrocks:
Yeah, that sounds like the best way forward. (or at least, how I would do that)

LyzardKing:
I only changed the data in ide.py. I'll check the others (games, web, and android).
Tell me if the changes in __init__.py and baseInstaller.py are good.
I added exec_path as None, and it is set in post_install

didrocks:
Ah last think, think about removal when executing umake <foo> <bar> --remove (you can do that in baseinstaller)

LyzardKing:
I wasn't sure about where to put the self.exec_path (init or baseinstaller). I had errors in the tests, but I realized the problems were mostly typos.
I'm cleaning it up now.

As for the removal, it worked with the version I uploaded 4 days ago. I'll check if it needs reworking.

LyzardKing:
I had left out the kwargs since exec_path will not be defined in the init (we don't have the instaled_path yet)

umake/frameworks/baseinstaller.py
```diff
@@ -114,6 +115,7 @@ def remove(self):
         if self.desktop_filename:
             with suppress(FileNotFoundError):
                 os.remove(get_launcher_path(self.desktop_filename))
+                os.remove(os.path.join(self.default_binary_link_path, self.desktop_filename.split('.')[0]))
```
LyzardKing: This is the removal of the link. We can create a get method, but other than that it works

umake/frameworks/baseinstaller.py
```diff
@@ -396,7 +398,8 @@ def decompress_and_install_done(self, result):
             UI.return_main_screen(status_code=1)
 
         self.post_install()
-
+        if self.desktop_filename:
```
LyzardKing: since we define self.exec_path only where it's needed, could we use that here instead of the desktop_filename?

didrocks: no, remember that more than one ide have "eclipse" as exec binary name? Maybe you're right, just create an @property in the base installer converting desktop filename to the property, (basically calling `self.desktop_filename.split('.')[0]`)

LyzardKing: I meant just for the condition, then we would still use desktop (or a property) to pass to add_exec_link

if self.exec_path:
            add_exec_link(self.exec_path, self.desktop_filename.split('.')[0])

didrocks: but you need to reuse the name for removing it. Hence reusing a property sounds useful :)

LyzardKing: Something like this?

@property
    def exec_link_name(self):
        if self.desktop_filename:
            return self.desktop_filename.split('.')[0]
        return None

didrocks: Sounds perfect :)

umake/tools.py
LyzardKing: Can I override the value for the binary directory defined in settings.py? I need it to point to the tmpfolder created in the small test

didrocks: yeah, try with @patch() or mock
(look at some of our unit tests examples using mock/patch overriding either a some of the tools.py method).

LyzardKing:
The small test works now

didrocks:
I think you should set `settings_module.DEFAULT_BINARY_LINK_PATH` to something like a temp dir (with a context manager so that the temp dir is directly removed). Otherwise, this looks good.

LyzardKing:
I fixed the conflicts and merged with upstream/master.
Isn't self.local_dir a temp dir? I intended to reuse that

didrocks:
oh, you're right, looking good then, did you run a test batch with this? (I'll just rely on you and can't double check right now ;))

LyzardKing:
I did. The tests work. I'll have another look tomorrow to finish the tests (at the moment the asserts are only on eclipse, lighttable, and vscode). But that should be simple enough

umake/frameworks/baseinstaller.py
didrocks: Maybe, we can even set:
if self.exec_link_name (to abstract from desktop_filename)

LyzardKing: Done

umake/frameworks/baseinstaller.py
```diff
@@ -59,6 +59,7 @@ def __init__(self, *args, **kwargs):
         self.desktop_filename = kwargs.get("desktop_filename", None)
         self.icon_filename = kwargs.get("icon_filename", None)
         self.match_last_link = kwargs.get("match_last_link", False)
+        self.exec_path = kwargs.get("exec_path", False)
```
didrocks: So, we can maybe change this for:
`self.exec_rel_path=`
Set that to the relative path (relative to installation path), and change that in each __init__() to add this key.

Then, we set self.exec_path base on self.install_path and self.exec_rel_path in `confirm_path()`

That way, you can remove in post_install the duplication of setting the self.exec_path, which can be error prone

Making sense?

PS: do not forget to change the tests/large/test_baseinstaller.py as well to take that one into account ;) (with and without a desktop filename)

LyzardKing: That makes a lot more sense. I'll work on that tomorrow

LyzardKing:
The last commit works. To be sure I'm running the tests again.
I noticed that required_files_path points to the main executable. Could we reuse that instead of adding exec_rel_path?

didrocks:
Sure! please reuse it :)
(I guess android-studio should be one of them as well)

LyzardKing:
Ok, so I'll remove exec_rel_path, use requred_files_path (it only has one file anyway) and reset the condition to create the link if the .desktop file exists. Do you agree? (if we never have more binaries in it, required_files_path could simply be a single element, or am I missing something?)
so instead of required_files_path=[os.path.join("bin", "studio.sh")]
it would be required_file_path = os.path.join("bin", "studio.sh")

LyzardKing:
Could we need multiple binaries in the future?

umake/tools.py
```diff
@@ -337,6 +337,14 @@ def create_launcher(desktop_filename, content):
         gsettings.set_strv("favorites", launcher_list)
 
 
+def add_exec_link(exec_path, destination_name):
+    bin_folder = settings.DEFAULT_BINARY_LINK_PATH
+    os.makedirs(bin_folder, exist_ok=True)
+    add_env_to_user("Umake symlink", {"PATH": {"value": bin_folder}})
```
didrocks: nitpick: I would have call it "Ubuntu Make binary symlinks"

didrocks:
I think there is a high change that one framework creates multiple binaries, so keep it as a list (there will be a list of desktop files then IMHO)

But in the implementation, for now, in @property, just take the first one from the list and add a symlink to it is there is a desktop file. We'll revisit later on if refinements are needed. Making sense?

LyzardKing:
ok. So, remove the exec_rel_path and use the first in required_files_path. Right?

didrocks:
yep! :)

LyzardKing:
Done. I'm running all the test just to be sure everythong works.

LyzardKing:
All the medium tests work. I added the asserts in ide, games, android-sdk

didrocks:
you meant android-studio? (you didn't push it yet ;))

Maybe web as well?

LyzardKing:
Yes, android-studio. I pushed it now.

didrocks:
I guess you didn't push the web commit (like firefox-dev) ;)

Basically grep desktop_filename to ensure we have everything covered?

LyzardKing:
That should be all the desktop_filename instances being used

didrocks:
Rock on! Merging it then. Thanks again for this (once again) excellent work :)

PR #241, opened 2016-02-01
Lighttable support

LyzardKing:
added tests, mock binary, certificates for the vm
Tests are working on a rebuilt image.
I used github.com releases (https://github.com/LightTable/LightTable/releases), since the files are downloaded from there(and the md5sum is on the same page)


didrocks:
Excellent! All looks good to me. Merging right away :)
Nice work as usual!

PR #242, opened 2016-02-02
Add JetBrains Datagrip

oijazsh:
Adds JetBrains DataGrip installer to ubuntu-make.

Medium and large test passes.


LyzardKing:
Looks great. It's like the other jetbrains installers. The tests are running so I think we can merge.

didrocks:
Thanks guys! Even better when it's spotless PR like this :)

PR #244, opened 2016-02-05
Fix symlink overwrite if exists

LyzardKing:
On reinstall, the symlink is removed, to ensure it points to the correct binary.


LyzardKing:
I noticed that I introduced a bug with the symlink merge.
I think the best thing to do is remove the previous link and replace it, in case the binary has moved in the archive.

LyzardKing:
or we could simply:

suppress FileNotFound:
    os.remove(file)
os.symlink(..., file)

didrocks:
That was the best pick! Looking good :)

PR #246, opened 2016-02-08
Add updated openjdk

LyzardKing:
Added a check in requirements_handler to grab the latest openjdk.

Added a check in the RequirementsHandler, where the requirements are verified.
If in a framework we put openjdk-latest we check:
  if in apt.cache we have openjdk-8 install that
  if not, check for openjdk-7.
To maintain this we just need to add new versions of openjdk when they get released in ubuntu.

The ideal thing would be to add a condition to avoid a particular version, say for example if eclipse wasn't working with openjdk-9.

Alternatively, we could keep default-jdk, and say that if another "java" is installed, it should skip it.
So if nothing is installed, umake would install the default (openjdk-7 at the moment). This could be done in the same place as my commit. (if "javac" exists as a command, remove "default-jdk" from the bucket)

Any ideas?

### DIFF
--- a/umake/frameworks/android.py
+++ b/umake/frameworks/android.py
@@ -79,7 +79,7 @@ class AndroidStudio(umake.frameworks.baseinstaller.BaseInstaller):
     def __init__(self, category):
         super().__init__(name="Android Studio", description=_("Android Studio (default)"), is_category_default=True,
                          category=category, only_on_archs=_supported_archs, expect_license=True,
-                         packages_requirements=["openjdk-7-jdk", "libncurses5:i386", "libstdc++6:i386", "zlib1g:i386",
+                         packages_requirements=["openjdk-latest", "libncurses5:i386", "libstdc++6:i386", "zlib1g:i386",
                                                 "jayatana"],
                          download_page="https://developer.android.com/sdk/index.html",
                          checksum_type=ChecksumType.sha1,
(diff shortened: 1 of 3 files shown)


didrocks:
Excellent work! However, I'm unsure about this work TBH.

I guess we want to ensure that we have an excellent developer story, and for IDEs/fraemworks, I saw a lot of times frameworks depending on an exact openjdk/java version because they don't really trust upstream.

On those, depending on "latest" sounds tricky, and so we would bump user's requirement without really knowing (apart from "we can start the app") the real status.
Thoughts on that? Do you think it worthes testing and risking running in an uncontrolled environment?

On the code itself:
- looks good, obviously miss some tests :)
- I would prefer we iterate (like try even 10-11 maybe to 15…) without fixing too much the version (from top to bottom) so that we don't really have to worry about updating.
- ack if we go that path to remove requirements if javac --version matches on of our expected version number (and that way, we let the liberty for people not having openjdk installed)

LyzardKing:
I'm not too sure about it either. But I manually install openjdk-8 anyway, so I end up having two different versions.
Of the two ideas the second could be better, to avoid issues. If there is already a minimum java version (openjdk-7 or defaul-jdk) installed, remove java from the requirements.
So if I manually install openjdk-8, calling umake after will not  install openjdk-7
I thought that having the possibility to say "latest-java", or being able to specify "openjdk-7" could have solved that. We keep "latest" if there are no issues, otherwise we change to a specific version.
It's just an idea, since I end up having two versions.
Let me know, and hopefully we can have other opinions on this.
I realize it would be a big change, so it's better to be sure

didrocks:
Yeah, I'm afraid still at unseen side-effects with running newest version that the one upstream supported.

We can as well go this way, and revert to a specific openjdk version if there is any complain. So I still think your feature can be valuable. I would be opened to get more opinions on this as well.

I think you might add the unit tests with different package configuration in the requirement handlers anyway to ensure we have all this covered.

(on another note, I'm going to cut a release tomorrow, I think I would prefer without that branch in ;))

LyzardKing:
I had no intention of proposing this for the next release, possibly not even the one after.
This has to be discussed a lot more. Of the things that have java as a dependency I only use eclipse and android-sdk on occasions, so I'm not qualified to say what works with different java versions.
I changed all of the dependencies in this branch just to test them.
I wanted mainly to see if this was something that others wanted.
I can close the pr and open an issue, to see if others reply with different opinions.

didrocks:
I just added the "question" label, to mark it as such, I think it's good enough :)

PR #251, opened 2016-02-18
Logic OR in dependency list

LyzardKing:
Add the possilibity to specify a dependency on one of n packages.
At least one of them has to be installed for the dependency to be met

Example. The Eclipse versions could depend on openjdk-8-jdk or openjdk-7-jdk (or even oracle-jdk).
All we need to do is specify an order.
The dependency on packages `"A | B | C"` would: ("|" being the or)

- check if one of them is installed
- if not install the first available.

### DIFF
--- a/tests/small/test_requirements_handler.py
+++ b/tests/small/test_requirements_handler.py
@@ -496,7 +496,7 @@ class TestRequirementsHandler(DpkgAptSetup):
                     os.path.join(self.dpkg_dir, "status"))
         self.handler.cache.open()
         self.assertTrue(self.handler.is_bucket_installed(test_bucket))
-        self.assertTrue(test_bucket == ['testpackage'])
+        self.assertEqual(test_bucket, ['testpackage'])
 
     def test_or_option_second_installed(self):
         test_bucket = ["testpackage0 | testpackage | testpackage1", "testpackage2"]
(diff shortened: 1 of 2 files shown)


didrocks:
That looks correct! However, the "if not install the first available" is hidden in the global logic. I'm afraid we break it without realizing.

I suggest that we get more global tests case (the one installing the package in the chroot), with:
* The 2 alternatives available -> check only the first one is installed
* Only the 2nd alternative available -> check that the second is installed

Otherwise, the diff provides an elegant solution for me, avoiding duplication, good work!

LyzardKing:
I added in the tests a check of the new bucket, created by removing the items with " | ", adding the correct one instead.

didrocks:
Looking good on the tests itself!
However, I'm unsure about the assertion, I would prefer an assertEquals(). That way, if the test fail, instead of "False is not True", we will see the 2 list content which differs.

LyzardKing:
You're right :)
Fixed now with assertEqual

didrocks:
Looks good! feel free to merge it now :-)

LyzardKing:
Thanks! :)
Now I think we should have a "guideline" on what the defaults should be.
For instance, the eclipse installers could depend on:
`["openjdk-8-jdk | openjdk-7-jdk | [the oracle version?]"]`

or we could keep openjdk-7-jdk as the first one.

I'll merge this  now.

It would be interesting, if possible, to know what versions people are using.

didrocks:
yeah, basing on data would be the best! I would say, let's try to keep the same base version as long as it's not required otherwise for upstream, for the sake of not creating issue on some configurations and not on others

LyzardKing:
Sounds perfect! I'll merge this for now, and the next thing will be to change the dependencies to:
`["openjdk-7-jdk | openjdk-8-jdk"]` for the various eclipse/jetbrains/android(?).
I'll check to see if they support openjdk-8 first to be sure.
Then if somebody uses the oracle version we can add that as well.

didrocks:
sounds like a good plan :)

PR #252, opened 2016-02-20
Remove openjdk from datagrip requirements.

oijazsh:
I overlooked that Datagrip is bundled with a custom jdk. A system installed java is hence not required.

### DIFF
--- a/umake/frameworks/ide.py
+++ b/umake/frameworks/ide.py
@@ -457,7 +457,7 @@ class DataGrip(BaseJetBrains):
                          only_on_archs=['i386', 'amd64'],
                          dir_to_decompress_in_tarball='DataGrip-*',
                          desktop_filename='jetbrains-datagrip.desktop',
-                         packages_requirements=['openjdk-7-jdk', 'jayatana'],
+                         packages_requirements=['jayatana'],
                          icon_filename='product.png')
 
 


LyzardKing:
Do you know whether this will continue in the future releases of it it's a one time thing?
Normally intellij only has custom-jdk available in some early access versions, as far as I know.
Also, a question for @didrocks, does jayatana need openjdk-7-jre installed? I suppose if we remove the jdk we still need the jre, or am I wrong?

oijazsh:
So far all four DataGrip releases (1.0.0 to 1.0.3) have had a jdk bundled. That said, I am sure not whether that will continue to be the case. Their download page lists openjdk 1.7+ as a requirement (as well as GNOME/KDE funnily enough) but looking at the datagrip.sh script, it appears not to use the system java if the bundled one is available.

For what it is worth, the bundled java is `openjdk version "1.8.0_76-release"`.

I myself do not use unity as a desktop environment so I can not comment on jayatana, other than that its dependencies are 
```
libc6, libdbusmenu-glib4, libglib2.0-0, libx11-6, gir1.2-glib-2.0, gir1.2-dbusmenu-glib-0.4
```

Tinche:
JFTR, recent JetBrains software blocks jayatana explicitly anyway. For example, PyCharm 5.0.4. Professional, pycharm.sh:

```
if [ -n "$JAVA_TOOL_OPTIONS" -a "$JAVA_TOOL_OPTIONS" != "${JAVA_TOOL_OPTIONS%-javaagent*jayatanaag.jar*}" ] ; then
  export _ORIGINAL_JAVA_TOOL_OPTIONS="$JAVA_TOOL_OPTIONS"
  JAVA_TOOL_OPTIONS="${JAVA_TOOL_OPTIONS%-javaagent*jayatanaag.jar*}${JAVA_TOOL_OPTIONS#*jayatanaag.jar}"
fi
```

I have jayatana installed but PyCharm skips it anyway, apparently because of bugs.

LyzardKing:
Ok.Thank you both!
@oijazsh. If that's the case we can remove the dependency. We'll reintroduce it later if they change the distributed version.
@Tinche. We can also remove jayatana as a dependency then.
It confirms what you said here
https://blog.jetbrains.com/idea/2016/01/intellij-idea-15-0-3-update-is-available/
In the comments they explain why it's disabled:
"The problem is that the integration does not work well with the latest IDEA, especially for multi-project scenario"

didrocks:
+1 @LyzardKing.

Let's do that, remove jayanata anyway. I hoped that upstream would fix the issue we had at some point, but it seems this never happened. The version in the distro is harmless (basically didn't enable the env variable as in the past), but yeah, I guess there is no need to have the dep anymore.

Issue #253, opened 2016-02-23
Error installing unity3d


Alvarz:
I guys, keep doing the great job!

I got a fresh install of ubuntu  14.04.3 and when I try to install unity3d throught umake, it gave that error:

'ERROR: installArchives() failed
Subprocess output: E:Reverse conflicts early remove for package 'libglapi-mesa:amd64' failed'

then I went to synaptic to see whats going on and I could notice I had installed  libglapi-mesa-lts-virvid:amd64. 

I made a search about it and ubuntu 14.04.3 ship with libglapi-mesa-lts-virvid:amd64 so it's there by default, in any case if I try to force uninstall it and then install libglapi-mesa:amd64,it try to delete my whole system, even ubuntu.desktop.

What can I do?

LyzardKing:
This has to do with the lts-enablement stack. Updated versions of mesa that are added with a different name (ex libglapi-mesa-lts-vivid, or libglapi-mesa-lts-wily).
I think we can add them as possible dependencies. On non lts it should find the normal version.

Alvarz:
Thanks for the reply!!

So should I wait for the next update or should I install it throught the deb package on unity3d website?

didrocks:
It's looking good to me. However, I'll appreciate as well testing from someone on 14.04. @LyzardKing: do you mind running the testssuite in xenial, wily and trusty?
`./runtests vm <your_general_test_command>`. It should tell/yell at you and indicate how to build the vm when needed! Tell me if that doesn't work for you ;)

@Alvarz: do you mind using that branch? It's pretty easy (install git first! ;)), just:
```
git clone https://github.com/LyzardKing/ubuntu-make
cd ubuntu-make
git checkout unity-3d_lts_mesa
bin/umake games unity3d
```
That way, you have access on that branch to the latest code and can keep us posted if your issue is fixed that way! Thanks ;)

Alvarz:
@didrocks  yeah I would love to help on this project. 
Should I first uninstall my umake to test this branch?

LyzardKing:
No need to uninstall. if you follow the commands in the previous comment you're good to go

LyzardKing:
@didrocks The ./runtests vm test.. returns "adt-run isn't installed.".
Am I missing something obvious? (Might be :))

LyzardKing:
@didrocks I figured out what was missing, I needed some packages on my system.
I'll let you know if I get somewhere.
@Alvarz If you need help testing let me know!

[...]

PR #254, opened 2016-02-23
Unity3D on lts mesa updates

LyzardKing:
Added as possible dependencies mesa-lts
Closes: #253 

I'm not running ubuntu 14.04. If somebody could test this to see if we need to change other dependencies it would be good.

On ubuntu non lts it should stop at the first version (non lts); on 14.04 it should stop at the first available one. A problem could arise if there is no version installed. Is that possible?

### DIFF
--- a/umake/frameworks/games.py
+++ b/umake/frameworks/games.py
@@ -121,14 +121,15 @@ class Unity3D(umake.frameworks.baseinstaller.BaseInstaller):
                          need_root_access=True,
                          # Note that some packages requirements essential to the system itself are not listed (we
                          # don't want to create fake packages and kill the container for medium tests)
-                         packages_requirements=["gconf-service", "lib32gcc1", "lib32stdc++6", "libasound2", "libcairo2",
-                                                "libcap2", "libcups2", "libfontconfig1", "libfreetype6", "libgconf-2-4",
-                                                "libgdk-pixbuf2.0-0", "libglu1-mesa", "libgtk2.0-0",
-                                                "libgl1-mesa-glx | libgl1-mesa-glx-lts-vivid | libgl1-mesa-glx-lts-wily",
-                                                "libnspr4", "libnss3", "libpango1.0-0", "libpq5", "libxcomposite1",
-                                                "libxcursor1", "libxdamage1", "libxext6", "libxfixes3", "libxi6",
-                                                "libxrandr2", "libxrender1", "libxtst6",
-                                                "monodevelop"])  # monodevelop is for mono deps, temporary
+                         packages_requirements=[
+                             "gconf-service", "lib32gcc1", "lib32stdc++6", "libasound2", "libcairo2",
+                             "libcap2", "libcups2", "libfontconfig1", "libfreetype6", "libgconf-2-4",
+                             "libgdk-pixbuf2.0-0", "libglu1-mesa", "libgtk2.0-0",
+                             "libgl1-mesa-glx | libgl1-mesa-glx-lts-vivid | libgl1-mesa-glx-lts-wily",
+                             "libnspr4", "libnss3", "libpango1.0-0", "libpq5", "libxcomposite1",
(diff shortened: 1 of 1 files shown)


Alvarz:
I still have the error, is this already on the main branch?

LyzardKing:
Not yet. I would prefer to test this, before merging. I'll setup a 14.04.4 vm and try it there. If you want you can test the branch.

LyzardKing:
@Alvarz Did you test the branch?

Alvarz:
@LyzardKing [#253](https://github.com/ubuntu/ubuntu-make/issues/253) here is the result of the tes, sadly it did'nt work for me

LyzardKing:
@Alvarz Test the latest version. This should work.
I changed the order so it will install the latest mesa version, not the first available (the old ones remain available)
Let me know

Alvarz:
@LyzardKing  Yeah pal It's worked like a charm!!! I have already unity thanks to you!.

So this brach actually works on ubuntu 14.04.3

LyzardKing:
Great! I was really counting on this to work.
I wasn't too sure it would though. ;)
I'll merge then, so this can be in the next release.

LyzardKing:
@Alvarz 
Sorry, there was a typo in the code. It should not make a difference, but could you test it once more?

Alvarz:
Guys I think it had a big issue. It install unity 3d but after restart the SO stop working, dont boot, anything. When i hsd been looking throught the terminal i have many erros with libgl1-mesa-glx, libgl1-mesa-grid an all xserver* packages i  dont know if this was a coincidence or if this cause the error. I recommend do more tests on virtual machines

Alvarz:
Sorry the typing errors Im on my mobile and its on spanish

LyzardKing:
I'm sorry! I did test it in a vm without issues, but apparently it was set
up different from your system.
I'd say the problem is with the version of libgl1-mesa-glx we end up
selecting being different from the one in your system.
In your case nothing should have happened, since one of the supported
versions was already installed.

Alvarz:
Yeah, I've already fixed my system but unity3d stuck on starting, I dont
know if is cause a some packages issue, or if is unity 3d itself. I will
dig on it

LyzardKing:
I'd ask you to try the latest version (Fix typo), but I don't want to cause issues to your system.
Could you let me know what version of the kernel and enablement stack? 
Is your system a clean(ish) version of ubuntu 14.04.3?
So I can try and replicate a similar vm

LyzardKing:
After some testing, I arrived at a possible conclusion.
Since the different versions of lts stack updates are all in 14.04 apt cache, they can conflict with one another.
I believe it's better to fail the install than to break a system.
None of the two should happen anyway, if we cover all possibilities.
So we should set the order from oldest to newest.
If a version is already installed (should be the case in ubuntu desktop) -> do nothing;
if no version is installed for some reason(I don't think it's possible) -> install the first (non-lts updated) or fail.
I think this is the best thing we can do for now.

LyzardKing:
Yes. It should not be possible to have an ubuntu install with X, that is missing libgl1-mesa-glx from what I gather. So there should be no problems.
I'm still not sure about what the issue was on your system.

didrocks:
I guess we are good then! +1 from my standpoint :)

LyzardKing:
It should work. I'll test this on clean trusty installs in the next days to
be sure we 'we not breaking anything.
If those tests pass I'll merge.

LyzardKing:
@Alvarz Did the last version work? Or did it cause any problem?
I'm trying to figure out what went wrong before merging

LyzardKing:
Ok. xorg depends on libgl1-mesa-glx.
This means that a version of libgl1-mesa-glx has to be installed on your system.
This solves the doubt I had.
I'll merge now.

Alvarz:
Hello Galileo I have been installed it throught the deb package from
unity3d website without any issue.

PR #245, opened 2016-02-07
Add Superpowers framework

Tankypon:
[Superpowers](http://superpowers-html5.com/index.en.html) is a 2D+3D game maker. I think it was a good time to add it to umake, now that superpowers is free and open-source.

PR #256, opened 2016-02-24
Adds the kotlin language compiler

oijazsh:
This PR adds the kotlin language compiler.


umake/frameworks/kotlin.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def __init__(self):
+        super().__init__(name="Kotlin", description=_("The Kotlin Programming Language"), logo_path=None)
+
+
+class KotlinLang(umake.frameworks.baseinstaller.BaseInstaller):
+
+    def __init__(self, category):
+        super().__init__(name="Kotlin Lang", description=_("Kotlin language standalone compiler"),
+                         is_category_default=True, category=category, packages_requirements=["default-jre"],
```
LyzardKing: instead of "default-jre" put "openjdk-7-jdk | openjdk-8-jdk"
This way it will succeed if at least one of the two is installed

LyzardKing:
I was working on the same thing :)
You need to add the supported archs;
Put the kotlin folder in a folder "JetBrains" in the mock server;
Then there may be an issue with the "latest" page on github releases(with the redirect)

umake/frameworks/kotlin.py
oijazsh: Noted. Thanks :)

oijazsh:
Moved the kotlin folder to its right place. Sorry about that.

Is kotlin restricted to only i386 and amd64 architectures? I could not find any reference regarding supported architectures so I did not restrict the list.

I realize the issue with the redirect - as it is written the medium test is bypassing the redirect. Ideally, our mock 'latest' page should only be a redirect.

LyzardKing:
I'm not sure about the requirements, but without the supported architectures the medium tests fail with an invalid choice.
Is this not an issue for you? (It honestly could be my system)
The large test works.
I disagree with the solution you adopted for the medium test, since it does not represent reality. But  it's a larger issue that we need to solve independently from this.

LyzardKing:
I'll try to come up with a better solution for these type of redirects tomorrow. For now I think we can use your solution.
If the medium test works for you I'll have a look on my end.
A part form that good work!

tests/large/test_kotlin.py
```diff
@@ -0,0 +1,68 @@
...
+    def test_default_kotlin_install(self):
+        """Install Kotlin from scratch test case"""
+        if not self.in_container:
+            self.example_prog_dir = tempfile.mkdtemp()
+            self.additional_dirs.append(self.example_prog_dir)
+            example_file = os.path.join(self.example_prog_dir, "hello.kts")
+            open(example_file, "w").write(self.EXAMPLE_PROJECT)
+            compile_command = ["bash", "-l", "-c", "kotlinc -script {}".format(example_file)]
+        else:  # our mock expects getting that path
+            compile_command = ["bash", "-l", "kotlinc /tmp/hello.kts"]
```
LyzardKing: needs to be `kotlinc -script`

oijazsh:
The scala framework does not appear to have a supported archs list. Does that one work for you?

And yes, I agree that a better approach should be worked out for the redirects required in our medium tests before this is merged.

oijazsh:
The medium does now fail with an invalid choice - that goes away if I replace "openjdk-7-jre | openjdk-8-jre" with "default-jre" again however. I'll give it a look to see if I have missed anything.

LyzardKing:
Found it. The issue is that the docker image has default-jre as a package but not the other ones. I'd keep openjdk-7 and openjdk-8 as dependencies, so we can explicitly set the supported ones (default-jre and default-jdk can be different in different ubuntu releases.)
We can set those then in the docker image.

This is something that should be discussed.
@didrocks , what do you think?

didrocks:
@LyzardKing Agreed, let's get them in the default docker image for frameworks that needs it. You just need to modify `docker/create_packages.sh` as you probably already know). I'm happy to push it back to dockerhub the modified images.

Btw, great discussions guys! Awesome to see the current dynamic in the various PR from the Ubuntu Make community \o/

LyzardKing:
@didrocks  Done! Now the image can be rebuilt. As soon as that's done we're good to go.
Also, @Tinche had an idea to solve the issues with the redirects for github. We can use the github api. 
(https://github.com/ubuntu/ubuntu-make/pull/245#issuecomment-188477994)

oijazsh:
Changed it to utilize the github api. Large test passes. Medium tests likely needs new certificates now.

LyzardKing:
If you want to build a key for the server, you can have a look at tests/tools/local_server.py for info

LyzardKing:
Also you need to move the downloads (with the zip file) to the github.com folder (not api.github.com).
Then change the latest file to point to that (like the original page)
Let me know if you can do that.

LyzardKing:
As soon as this is completed I can upload the change to the other frameworks that rely on github.

oijazsh:
I have made the changes but have not been able to check the medium test. Due a network issue, `docker build .` fails to fetch the required repos. I will try to get that resolved and run the test tomorrow if possible.

If the test passes I'll rebase the commits to be neater if required.

LyzardKing:
try uncommenting the line `DOCKER_OPTS="--dns 8.8.8.8 --dns 8.8.4.4"` in /etc/default/docker
I had that problem some time ago.

LyzardKing:
It wasn't showing me the right diff earlier. Perfect!
I'll run a few tests and then merge if they pass

LyzardKing:
The medium tests fail with a timeout

LyzardKing:
also, you have to rename the folder in github.com/ to JetBrains

oijazsh:
My apologies, renamed.

LyzardKing:
No problem :)
I'm just trying to figure out what's not working in the medium test

LyzardKing:
Ok. I'm really not sure why, but the medium tests now work. The only thing I changed is in medium/test_kotlin.py:
`self.hosts = {443: ["api.github.com", "www.github.com"]}`
to
`self.hosts = {443: ["api.github.com", "github.com"]}`

If you can change that then we should be good to go.

oijazsh:
Thanks for noticing that. Changed it.

Also tidied up the commit history now that it is ready for merge.

LyzardKing:
All the tests now work. The docker image needs to be rebuilt, but it's good.
Merging now.

Issue #259, opened 2016-03-08
Error in ela accept and can't install visual code new release


overbid:
Can't run Visual code and got this message

```
Hello from Seattle.

Follow @code





window.twttr=(function(d,s,id){var js,fjs=d.getElementsByTagName(s)[0],t=window.twttr||{};if(d.getElementById(id))return t;js=d.createElement(s);js.id=id;js.src="https://platform.twitter.com/widgets.js";fjs.parentNode.insertBefore(js,fjs);t._e=[];t.ready=function(f){t._e.push(f);};return t;}(document,"script","twitter-wjs"));



Privacy
Terms of Use
License



&copy; 2016 Microsoft







































{
"@context" : "http://schema.org",
"@type" : "Product",
"name" : "Visual Studio Code",
"logo": "https://code.visualstudio.com/Content/images/vscode.ico",
"url" : "https://code.visualstudio.com",
"potentialAction": {
"@type": "SearchAction",
"target": "https://code.visualstudio.com/Search?q={search_term_string}",
"query-input": "required name=search_term_string"
},
"sameAs" : [
"https://en.wikipedia.org/wiki/Visual_Studio_Code",
"https://www.facebook.com/visualstudio",
"https://twitter.com/code",
"http://blogs.msdn.com/b/vscode"
]
}
```

LyzardKing:
They slightly changed the license page. I'll fix it now.

overbid:
And them change run file from Code to code, too.

LyzardKing:
Done. I fixed and ran the tests and now it works.
@didrocks  I'd leave the issue open until we release a version that has this fix

overbid:
I try to pull request but @LyzardKing, you fix it very quick.

PR #261, opened 2016-03-11
Add more test for vscode

overbid:
from https://github.com/ubuntu/ubuntu-make/issues/259


didrocks:
Thanks a lot for spotting that one! All good, +1 :)

PR #269, opened 2016-03-20
Fix medium tests

LyzardKing:
The medium tests all work now.
The next thing to do is look at the large tests.
Mostly they should work, but I can't get the vm to run properly.

### DIFF
--- a/umake/frameworks/baseinstaller.py
+++ b/umake/frameworks/baseinstaller.py
@@ -59,10 +59,9 @@ class BaseInstaller(umake.frameworks.BaseFramework):
         self.desktop_filename = kwargs.get("desktop_filename", None)
         self.icon_filename = kwargs.get("icon_filename", None)
         self.match_last_link = kwargs.get("match_last_link", False)
-        self.exec_rel_path = kwargs.get("exec_rel_path", None)
         for extra_arg in ["download_page", "checksum_type", "dir_to_decompress_in_tarball",
                           "desktop_filename", "icon_filename", "required_files_path",
-                          "match_last_link", "exec_rel_path"]:
+                          "match_last_link"]:
             with suppress(KeyError):
                 kwargs.pop(extra_arg)
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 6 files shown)


tests/medium/__init__.py
```diff
@@ -31,7 +31,7 @@ class ContainerTests(LoggedTestCase):
     """Container-based tests utilities"""
 
     DOCKER_USER = "user"
-    DOCKER_TESTIMAGE = "didrocks/docker-umake-manual"
+    DOCKER_TESTIMAGE = "e3ff8df5bdd8"
```
didrocks: Please remove that one (and rebase so that it's not a double change :))

umake/frameworks/baseinstaller.py
```diff
@@ -137,6 +136,9 @@ def remove(self):
     def confirm_path(self, path_dir=""):
         """Confirm path dir"""
 
+        if self.desktop_filename:
```
didrocks: Why did you move that piece, you didn't comment it?

umake/frameworks/ide.py
```diff
@@ -77,7 +77,6 @@ def __init__(self, *args, **kwargs):
             kwargs["required_files_path"] = current_required_files_path
         download_page = 'https://www.eclipse.org/downloads/'
         kwargs["download_page"] = download_page
-        kwargs["exec_rel_path"] = "eclipse"
```
didrocks: I still find it being mentioned twice in baseinstaller.py:
```umake/frameworks/baseinstaller.py:        self.exec_rel_path = kwargs.get("exec_rel_path", None)
umake/frameworks/baseinstaller.py:                          "match_last_link", "exec_rel_path"]:
```

Did you remove them and I didn't notice it? If we are going to remove all of them, better to ensure it's not anywhere anymore :)

didrocks:
Here we go! Just some comments and questions to be answered before merging. Otherwise, good job! :)

tests/medium/__init__.py
LyzardKing: Ah yes... This was not supposed to be pushed. It was just to run the tests.

umake/frameworks/baseinstaller.py
LyzardKing: I moved this from the end of confirm_path to the start.
The medium baseInstaller test was failing.
The function hit a return before it could set the exec_path

umake/frameworks/ide.py
LyzardKing: It should be removed..
I made the changes on the wrong branch, so it might be that I forgot some when creating a new one

didrocks:
Merging this one as well now!

PR #271, opened 2016-03-22
Timeout on download

LyzardKing:
This is the timeout added to the download.
I had to add it in DownloadItem so it could be changed for the test.
I'm sure there is a better way to achieve this.

The ideal thing would be to change only lines 133, 134, 147, 148 in download_center.py,
and in test_download_center.py force the download to stop immediately.
That would represent better a real scenario.

### DIFF
--- a/tests/small/test_download_center.py
+++ b/tests/small/test_download_center.py
@@ -493,6 +493,14 @@ class TestDownloadCenter(LoggedTestCase):
         self.assertIsNone(result.fd)
         self.expect_warn_error = True
 
+    def test_download_timeout(self):
+        """we fail a download with timeout"""
+        filename = "simplefile"
+        url = self.build_server_address(filename + '-timeout')
+        request = DownloadItem(url, None)
+        DownloadCenter([request], self.callback)
+        self.assertIsNone(self.callback.call_args)
+
 
 class TestDownloadCenterSecure(LoggedTestCase):
     """This will test the download center in secure mode by sending one or more download requests"""
--- a/tests/small/test_download_center.py
+++ b/tests/small/test_download_center.py
@@ -495,12 +495,11 @@ class TestDownloadCenter(LoggedTestCase):
 
     def test_download_timeout(self):
         """we fail a download with timeout"""
-        filename = "simplefile"
-        url = self.build_server_address(filename + '-timeout')
-        request = DownloadItem(url, None)
-        DownloadCenter([request], self.callback)
-        self.assertIsNone(self.callback.call_args)
 
+        filename = "simplefile-timeout"
+        request = self.build_server_address(filename)
+        DownloadCenter([DownloadItem(request, None)], self.callback)
+        self.assertIsNone(self.callback.call_args)
 
 class TestDownloadCenterSecure(LoggedTestCase):
     """This will test the download center in secure mode by sending one or more download requests"""
--- a/tests/tools/local_server.py
+++ b/tests/tools/local_server.py
@@ -29,6 +29,7 @@ import ssl
 from . import get_data_dir
 import urllib
 import urllib.parse
+import time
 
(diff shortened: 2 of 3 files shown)


didrocks:
I wouldn't really go that path IMHO as it represents a global session timeout, isn't it? My approach would be rather to get the content as we do today, and timeout when we don't get any new data for x period of time (probably a minute?).

That way, we don't rely on any kind of short or long downloads and can be more flexible.

(FTR, I finally hope to be able to run the tests tomorrow once I'm back home, then, we freeze master for a release on Tuesday and can merge those kind of new features :)

LyzardKing:
I might have gotten this wrong, but what I intended to do is add a timeout to the requests.get() method.
This should timeout after x seconds (I used 10 seconds just to test).
This should not be a global session timeout, only related to the get

didrocks:
I may not have it correctly then, sorry for it! I need to look closer it seems :)

Mind pushing it to 60s then? (Like, if you have a real network issue), and fix the test that is failing on travis?

tests/small/test_download_center.py
```diff
@@ -493,6 +493,20 @@ class WrongChecksumType(Enum):
         self.assertIsNone(result.fd)
         self.expect_warn_error = True
 
+    def test_download_timeout(self):
+        """we fail a download with timeout"""
+        filename = "simplefile"
+        url = self.build_server_address(filename)
+        request = DownloadItem(url, None, timeout=0.0000001)
```
didrocks: This seems a little bit racy, I would rather in the server add a real timeout issue.

Trigger that special behavior on <filename>-timeout for instance!

LyzardKing:
The issue in travis is that I had to force the timeout to a very low number to get it to fail in the test.
I guess it's not low enough when it runs in travis.
That's why it would be better to find another way to test it.
I'll push the 60s change

didrocks:
hence my new suggestion for the test ;)

LyzardKing:
Ah yes! I didn't see that comment :)
That would be a lot better. We could remove the timeout from DownloadItem and set it directly in the get.

LyzardKing:
How can I force a timeout on the server then?

didrocks:
I guess that needs to be implemented in the fake server. Look at `tests/tools/local_server.py`. I guess you  need to add some magic after reading the doc of SimpleHTTPRequestHandler to see what to override to create this behavior (like sending 3 packets, and then hangs)

LyzardKing:
I'll have a look at the docs then.
The only change in download_center.py would then be:
line 132: `with closing(session.get(url, stream=True, headers=headers,
                                     cookies=cookies, timeout=60.0)) as r:`
We don't need to get timeout from anywhere else

didrocks:
agreed! I don't see any reason to wait frameworks to override it. At most, we can set it as a constant in settings.py

LyzardKing:
Does the test_download_center use the do_GET in local_server.py?
If it does we could override that, making it simply wait for longer than the timeout

didrocks:
I'm pretty sure it does, as this is what serves the content for this background test.

LyzardKing:
I'm trying to figure out how to do it.
At the moment I'm breaking all the other tests ;)

LyzardKing:
The test still has conflicts with the others. Other tests that have nothing to do with the timeout are failing with a timeout :(
I removed most of the exception since it was repeating the same as the future.exception in _one_done (download_center.py)

didrocks:
See, tests are useful! I'm sure this is covering up some corner cases! :)

LyzardKing:
I'm not sure, it might be something wrong I did.
Other tests seem to use the -timeout option I added.

LyzardKing:
I'm closing this PR since it would need a lot more work to fix the issues with the tests.
Also I don't really think it's needed at the moment.
We can always reopen it at a later date

PR #276, opened 2016-03-26
Fix large tests

LyzardKing:
Swift command changed (It doesn't need to compile a binary and run it)
Intellij reverted to closing java (I changed it in a previous commit, but reverting to this solution works)
Remove link in tests (The correct thing happens running umake -r, the test needed to match that)

The fix missing is the checksum for SpringToolsSuite


didrocks:
This looks good! Merging and retesting the failing tests.

I would have love a test in baseinstaller to fail to check that the added binary isn't properly removed, but that we can added afterwards.

Cheers!

PR #280, opened 2016-03-31
GitHub repo

LyzardKing:
Get the latest release versions with the github api.
Added to LightTable and Superpowers the api parses used for kotlin.
Fixed the mock server and the tests.


tests/medium/test_games.py
```diff
@@ -73,16 +73,8 @@ class SuperpowersInContainer(ContainerTests, test_games.SuperpowersTests):
...
 
     def setUp(self):
-        self.hosts = {443: ["github.com"]}
+        self.hosts = {443: ["api.github.com", "github.com"]}
         self.apt_repo_override_path = os.path.join(self.APT_FAKE_REPO_PATH, 'superpowers')
         super().setUp()
         # override with container path
         self.installed_path = os.path.join(self.install_base_path, "games", "superpowers")
-
-    def test_install_with_changed_download_page(self):
```
didrocks: As you are overriding `get_metadata_and_check_license`, I still think it's important to override the download page in games and ides to fake the API returned values (seeding wrong content basically).

didrocks:
Ok, just having this small comment (see in file), but other than that, it's good. Do not forget to uncomment the `return` on spring tools suite (both large/medium tests and the framework itself) so that you reenable the framework :)

WIth the change is taken into account and if the tests pass, +1 :) Thanks for looking at using the api, it seems way more robust to me!

LyzardKing:
The spring-tools-suite hasn't changed.
It still needs fixing

didrocks:
Sorry, I misread LightTable ;)

didrocks:
+1 for those changes. Merging as well and many thanks :)

LyzardKing:
Great!

PR #281, opened 2016-03-31
added openjdk-8-jdk as a dependency to eclipse and jetbrains

LyzardKing:
Added openjdk-8 to eclipse and jetbrains ides.
Just want to be sure of a few things:
1. Are we sure that there are no issues with jdk-8? (there shouldn't be)
2. should we reintroduce jayatana?
3. Do the eap versions of jetbrains ides have the custom jdk?

### DIFF
--- a/umake/frameworks/ide.py
+++ b/umake/frameworks/ide.py
@@ -183,7 +183,7 @@ class EclipseJava(BaseEclipse):
                          dir_to_decompress_in_tarball='eclipse',
                          desktop_filename='eclipse-java.desktop',
                          category=category, only_on_archs=['i386', 'amd64'],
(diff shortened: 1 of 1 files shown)


Tinche:
As noted before, JetBrains stuff doesn't need openjdk any more. Maybe remove it completely?

LyzardKing:
Thanks! That's what I wanted to know. :)
Is that true for every jetbrains release?
If so I'll change it.

Tinche:
PyCharm for sure (been using it), Idea Community Edition too (just
installed now to test).

didrocks:
Let's do that then! (on jayatana: until we have a better story and it's tested, let's not include it)

Issue #279, opened 2016-03-31
Xenial removed openjdk-7


tbroyer:
Xenial has no openjdk-7-jdk package [1] which will likely break most Java-based IDEs' installation that have `packages_requirements=['openjdk-7-jdk'],`

I notice that some others (Kotlin, Scala) have `packages_requirements=["openjdk-7-jre | openjdk-8-jre"],`, so they should probably be updated to `packages_requirements=["openjdk-7-jdk | openjdk-8-jdk"],` (assuming it works on Xenial where openjdk-7-jdk doesn't exist at all).

That said, would `java7-jdk` work? (the Oracle package from Webupd8 provides it for instance, and `java7-jdk` is provided by Java 8 packages such as `openjdk-8-jdk`) I suppose the problem with `java7-jdk` is that being a virtual package you cannot `apt-get install` it? (never tried, just an hypothesis) Maybe listing `java7-jdk` as an alternative would allow an Oracle (or other) JDK? (would detect an installed JDK and as a result wouldn't install an OpenJDK).

Related to #251 and b3005f7981d6a710d22a2d72a009ed769f1e13cb

[1] https://bugs.launchpad.net/ubuntu/+source/openjdk-7/+bug/1563986

LyzardKing:
Kotlin and scala have their own compiler, independent from the jdk.
For oracle java we'd need to come up with a better way, since it can be installed in different ways.

I thought that xenial still had openjdk-7.
We can fix the dependencies, are we aware of any issues with intellij/eclipse ides with java-8?

tbroyer:
> I thought that xenial still had openjdk-7.

Actually, I tried an upgrade this morning and, seeing it wanted to uninstall openjdk-7, I cancelled and looked up the reason (starting with http://packages.ubuntu.com/search?keywords=openjdk-7&searchon=names&suite=xenial&section=all, vs http://packages.ubuntu.com/search?keywords=openjdk-7&searchon=names&suite=wily&section=all)

Re. other JDKs, would adding `java7-jdk` as alternative work at least for the Webupd8 packages?

> are we aware of any issues with intellij/eclipse ides with java-8?

Given that umake downloads the latest versions, and java-7 has been EOL'd one year ago (Oracle JDK, not OpenJDK, but still), I'd bet all issues have been solved for a long time; not to mention that all those IDEs (obviously) support _writing_ (compiling, testing, debugging, refactoring, etc.) Java 8.
[...]

PR #282, opened 2016-04-01
Openjdk 8

LyzardKing:
I can't reopen the previous pr. (#281)
I added openjdk8 to eclipse and
removed openjdk as a dependency from the jetbrains ides.

### DIFF
--- a/docker/create_packages.sh
+++ b/docker/create_packages.sh
@@ -60,6 +60,7 @@ dpkg-scanpackages . /dev/null | gzip -9c > Packages.gz
 mkdir -p $repo_root_dir/eclipse
 cd $repo_root_dir/eclipse
 create_package openjdk-7-jdk
+create_package openjdk-8-jdk
 dpkg-scanpackages . /dev/null | gzip -9c > Packages.gz
 
 # stencyl deps
--- a/docker/create_packages.sh
+++ b/docker/create_packages.sh
@@ -56,11 +56,10 @@ create_package libstdc++6 i386
 create_package zlib1g i386
 dpkg-scanpackages . /dev/null | gzip -9c > Packages.gz
 
-# eclipse deps
-mkdir -p $repo_root_dir/eclipse
-cd $repo_root_dir/eclipse
-create_package openjdk-7-jdk
-create_package openjdk-8-jdk
+# rubymine deps
+mkdir -p $repo_root_dir/rubymine
+cd $repo_root_dir/rubymine
+create_package ruby
 dpkg-scanpackages . /dev/null | gzip -9c > Packages.gz
 
 # stencyl deps
--- a/tests/large/test_ide.py
+++ b/tests/large/test_ide.py
@@ -125,7 +125,7 @@ class IdeaIDETests(LargeFrameworkTests):
     def setUp(self):
         super().setUp()
         self.installed_path = os.path.join(self.install_base_path, "ide", "idea")
-        self.desktop_filename = 'jetbrains-idea.desktop'
+        self.desktop_filename = 'jetbrains-idea-ce.desktop'
         self.command_args = '{} ide idea'.format(UMAKE)
         self.name = 'Idea'
 
@@ -216,7 +216,7 @@ class PyCharmIDETests(IdeaIDETests):
     def setUp(self):
         super().setUp()
         self.installed_path = os.path.join(self.install_base_path, "ide", "pycharm")
-        self.desktop_filename = 'jetbrains-pycharm.desktop'
+        self.desktop_filename = 'jetbrains-pycharm-ce.desktop'
         self.command_args = '{} ide pycharm'.format(UMAKE)
         self.name = 'PyCharm'
 
@@ -231,7 +231,7 @@ class PyCharmEducationalIDETests(IdeaIDETests):
     def setUp(self):
         super().setUp()
         self.installed_path = os.path.join(self.install_base_path, "ide", "pycharm-educational")
-        self.desktop_filename = 'jetbrains-pycharm.desktop'
+        self.desktop_filename = 'jetbrains-pycharm-edu.desktop'
         self.command_args = '{} ide pycharm-educational'.format(UMAKE)
         self.name = 'PyCharm Educational'
 
--- a/tests/medium/test_ide.py
+++ b/tests/medium/test_ide.py
@@ -37,7 +37,7 @@ class EclipseJavaIDEInContainer(ContainerTests, test_ide.EclipseJavaIDETests):
     def setUp(self):
         self.hosts = {443: ["www.eclipse.org"]}
         # we reuse the android-studio repo
-        self.apt_repo_override_path = os.path.join(self.APT_FAKE_REPO_PATH, 'eclipse')
+        self.apt_repo_override_path = os.path.join(self.APT_FAKE_REPO_PATH, 'android')
         super().setUp()
         # override with container path
         self.installed_path = os.path.join(self.install_base_path, "ide", "eclipse")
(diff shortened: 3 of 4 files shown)


didrocks:
Two small comments:
Did you try eclise with openjdk8, is it working well?
Also, I think you need to remove the creation of the docker local repository for intellij ides (as we won't use it), and as well, from their medium tests requirements. Let's not accumate some cruft! :)

LyzardKing:
I've been using eclipse (only the Java version, but they should be the
same) with openjdk-8 for a while, without issues.
I'll clean up the docker repos now

LyzardKing:
Acutally, we were using the android repository, right?

didrocks:
oh, if so, ignore me :) Just remove the reference to it in the medium tests thus!

LyzardKing:
Done. I missed that before ;)
I added openjdk-8-jdk to the eclipse repository.
I'll just check if the eap releases we get via the api also have the custom jdk.

LyzardKing:
Apparently the eap versions have the custom java as well.

openjdk version "1.8.0_76-release"
OpenJDK Runtime Environment (build 1.8.0_76-release-b18)
OpenJDK 64-Bit Server VM (build 25.76-b18, mixed mode)

didrocks:
If the tests pass, +1 for me :)

LyzardKing:
The medium tests work.
I'm running a few large tests in the vm to be sure.

didrocks:
So, you confirm that all your VMs issues are fixed? If so, awesome \o/

LyzardKing:
They were yesterday, but you never know ;)
I'll let you know as soon as the test ends

LyzardKing:
The vm setup works with upstream(this repository), not with my fork.
I'll dig into that later. 
For now I tried running a few tests locally, eclipse and intellij idea work.

LyzardKing:
Ok. The vm issues are fixed now!
Eclipse works in the vm, as does idea.

LyzardKing:
Are we sure that the full jdk is included in the custom java release?
According to this only the jre is included. So at least IDEA should keep the jdk.
Pycharm doesn't need the jdk, only the jre to run it.
https://www.jetbrains.com/help/idea/2016.1/requirements-for-intellij-idea.html#d715320e221

tbroyer:
Copying my comment from #279 (sorry):

Just downloaded IDEA Community Edition (manually); it starts without problem (using its own bundled JRE), and you can open and create projects without problem.
Because you do not have a JDK configured though, it asks you whether you really want to create the project without JDK (or you can configure a JDK directly from the wizard, assuming you have a JDK installed already).

So, the JDK is not strictly required for IDEA, but of course if you do Java development (or a JVM-based language such as Kotlin or Groovy) then you'll need a JDK. I'd assume a developer is savvy enough to install a JDK himself (particularly one who already found and installed umake :wink: )

**update:** …and that would solve the "issue" of using an Oracle JDK and not OpenJDK.

LyzardKing:
Kotlin doesn't need the jdk.
We have to decide if we want to install things like jdk. If so we might
also want to install Ruby for rubymine.
If not we can install only the things needed to run the ide.
For eclipse we'd install only the jre.
The important thing is to have a consistent behaviour.

tbroyer:
Maybe Kotlin doesn't need a JDK, but IDEA asks you to configure one (maybe it could be a jre, I don't know)

I agree about being consistent, my opinion being to only install what's needed to run the IDE.

LyzardKing:
Then we leave all the jetbrains ides without dependencies,
and we remove jdk from eclipse, replacing it with the jre.
So the ide can run.
If someone wants to compile they install the jdk.
Does this sound good?

LyzardKing:
I added the jdk to the java ides (idea).
I also changed the name of the .desktop files to avoid conflicts (see issue #283 )

LyzardKing:
I think we should rely on the dependencies outlined in the website description.
In this case we should add python to pycharm (not really needed since it's already included in ubuntu);
and ruby in rubymine.
The requirement for ruby specifies version 1.8 or later. Since in trusty the version is 1.9 we're good.

didrocks:
As dicussed by through email with @LyzardKing, I think this is the most reasonable way forward.

didrocks:
That looks good to me, did you run the medium and large tests?

I think I'll cut a release tomorrow with this to get it fixed before the next days :)

LyzardKing:
The large tests work (in the vm).
The medium tests for RubyMine (on the repository rubymine with the 
package ruby) work.
The medium tests for PyCharm don't.
I added a repository (pycharm) with the packages: python, python3
When I run the test I get:
ERROR: [Errno 1] Operation not permitted
100% 
|########################################################################|
ERROR: Package requirements can't be met: [Errno 1] Operation not permitted

didrocks:
waow, interesting, this happens normally when it can't install (due to apt locking) those. Want me to have a quick look tomorrow or digging into it?

LyzardKing:
Sure! :)
It could very well be a simple thing I got wrong.
It should simply ignore everything, since python and python3 are already 
installed.
Could it be that we're overriding a repository for a package that is 
installed?

LyzardKing:
@didrocks The test works without the repository override.
Could it be that we are overriding a repository for packages that are already installed?

didrocks:
After debugging, the issue in python is quite different. The packaging on python version has some magic so that the default python package (`python*-default`) provides `python` and `python3`. Those depends then on a specific version: `python3.4` for instance.

When we override the repository by adding our own python fake package, it doesn't depend on any python version. There is some packaging logic (due to multi-arch) to then conflict and remove those packages.
This is what happens when I try to install this package in the container, with the pycharm repo:

```sh
$ sudo apt install python3
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following packages were automatically installed and are no longer required:
  at dctrl-tools diffstat distro-info-data dput gir1.2-glib-2.0
  hardening-includes help2man iso-codes libapt-pkg-perl libarchive-zip-perl
  libauthen-sasl-perl libautodie-perl libclass-accessor-perl libclone-perl
  libcommon-sense-perl libcurl3-gnutls libdbus-glib-1-2 libdigest-hmac-perl
  libdistro-info-perl libemail-valid-perl libencode-locale-perl liberror-perl
  libexporter-lite-perl libfile-basedir-perl libfile-listing-perl
  libfont-afm-perl libgirepository-1.0-1 libhtml-form-perl libhtml-format-perl
  libhtml-parser-perl libhtml-tagset-perl libhtml-tree-perl
  libhttp-cookies-perl libhttp-daemon-perl libhttp-date-perl
  libhttp-message-perl libhttp-negotiate-perl libio-html-perl libio-pty-perl
  libio-socket-inet6-perl libio-socket-ssl-perl libio-string-perl
  libio-stringy-perl libipc-run-perl libipc-system-simple-perl libjson-perl
  libjson-xs-perl liblist-moreutils-perl liblwp-mediatypes-perl
  liblwp-protocol-https-perl libmailtools-perl libnet-dns-perl
  libnet-domain-tld-perl libnet-http-perl libnet-ip-perl libnet-smtp-ssl-perl
  libnet-ssleay-perl libparse-debcontrol-perl libparse-debianchangelog-perl
  libperlio-gzip-perl libsocket6-perl libsub-identify-perl libsub-name-perl
  libtext-levenshtein-perl libtie-ixhash-perl liburi-perl libwww-perl
  libwww-robotrules-perl libxslt1.1 libyaml-0-2 lintian patchutils
  python-apt-common python3-apt python3-dbus python3-gi python3-lxml
  python3-pycurl python3-yaml strace t1utils unzip
Use 'apt-get autoremove' to remove them.
The following packages will be REMOVED:
  dh-python lsb-release python3-argcomplete python3-bs4 python3-chardet
  python3-debian python3-gnupg python3-magic python3-nose python3-pep8
  python3-pexpect python3-pkg-resources python3-progressbar python3-requests
  python3-setuptools python3-six python3-software-properties python3-urllib3
  python3-xdg ubuntu-minimal unattended-upgrades
The following packages will be upgraded:
  python3
1 upgraded, 0 newly installed, 21 to remove and 1 not upgraded.
Need to get 0 B/668 B of archives.
After this operation, 4831 kB disk space will be freed.
```
For obvious reasons, I don't let packages to be removed (what's after `The following packages will be REMOVED:`) in Ubuntu Make and prefer to error in that case.
The fix will be to have an optional toggle to set multi-arch same/any on packages and use the `any` on our fake python* package. I'm opening a bug to fix it later on.

Merging without this meanwhile (it works as python and python3 are installed by default in Ubuntu, and so, in the container)

didrocks:
@LyzardKing  Note for later: please try to separate fixes (like the rubymine addition and the openjdk-8 fix) on different PR :)

PR #287, opened 2016-04-06
Updated requirements for Netbeans to work in Xenial

fcole90:
Hi, I just updated the Netbeans requirements to make it work under Xenial, adding openjdk 8 as an alternative requirement.

### DIFF
--- a/umake/frameworks/ide.py
+++ b/umake/frameworks/ide.py
@@ -629,7 +629,7 @@ class BaseNetBeans(umake.frameworks.baseinstaller.BaseInstaller):
                          download_page="https://netbeans.org/downloads/zip.html",
                          dir_to_decompress_in_tarball="netbeans*",
                          desktop_filename="netbeans{}.desktop".format(flavour),
-                         packages_requirements=['openjdk-7-jdk', 'jayatana'],
+                         packages_requirements=['openjdk-7-jdk | openjdk-8-jdk'],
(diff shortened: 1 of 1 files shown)


umake/frameworks/ide.py
```diff
@@ -629,7 +629,7 @@ def __init__(self, category, flavour=""):
                          download_page="https://netbeans.org/downloads/zip.html",
                          dir_to_decompress_in_tarball="netbeans*",
                          desktop_filename="netbeans{}.desktop".format(flavour),
-                         packages_requirements=['openjdk-7-jdk', 'jayatana'],
```
LyzardKing: Does netbeans disable jayatana? If not I'd leave that in

LyzardKing:
Great! This needed to be done.
Just one note on jayatana

umake/frameworks/ide.py
fcole90: jayatana is not really required for netbeans, and I'm removing it in line with the other IDEs to avoid possible problems with it (which IIRC is the reason because it's not enabled in IntelliJ and Eclipse)

LyzardKing: We removed it from the intellij ides recently because it was actually disabled in the exec sh file.
I'm not sure what's going on with jayatana.
Seems like removing it is the best thing to do now.

PR #289, opened 2016-04-08
Fix Android studio and sdk

### LINKED ISSUE #288: Android studio installation error
Hi,

After 2.0 [release](http://android-developers.blogspot.com.by/2016/04/android-studio-2-0.html) seems like something changed on download and installation flow.

When I use `umake android android-studio` and select destination folder console return an error:
`ERROR: We were expecting to find a license on the download page, we didn't.`

LyzardKing:
Added https in dl.google.com with cert for docker image
Medium tests work
fixes #288


umake/frameworks/android.py
```diff
@@ -71,6 +71,8 @@ def parse_download_link(self, tag, line, in_download):
 
         if url is None and sha1sum is None:
             return (None, in_download)
+        if url and url.startswith("//"):
```
LyzardKing: This is needed because the line with the url is:
`href="//dl.google.com/android/android-sdk_r24.4.1-linux.tgz">android-sdk_r24.4.1-linux.tgz</a>`

didrocks: Ahah, gotcha! :)

didrocks:
Excellent! Looks good to me :) Thanks again for your hard work there. merging.

PR #295, opened 2016-04-18
Sublime text

LyzardKing:
Added support for sublime-text version 3.
The large test works.
The medium tests works after rebuilding the docker image.


tests/medium/test_ide.py
```diff
@@ -411,6 +411,27 @@ def test_install_with_changed_download_page(self):
...
+    TIMEOUT_START = 20
+    TIMEOUT_STOP = 10
+
+    def setUp(self):
+        self.hosts = {443: ["sublimetext.com", "download.sublimetext.com"]}
+        super().setUp()
+        # override with container path
+        self.installed_path = os.path.join(self.install_base_path, "ide", "sublime-text")
+
+    def test_install_with_changed_download_page(self):
```
didrocks: You don't need this one as you are using the base parser in BaseInstaller. There is already a test in BaseInstaller to ensure we support bad page parsing :)

didrocks:
Perfect! Just look at my comment on the undeeded changed download page test that you can remove. Otherwise, all good!

didrocks:
Let's add it to the next release (after the current bug-fixing one), shall we? Otherwise, it looks +1 for me :)

Thanks again!

LyzardKing:
Sounds great.
I'm looking at a fix for arduino ide.
It should just be fixed by removing the dependency on openjdk(the jre is
bundled), but the test is failing now.
After that everything should work on 16.04

LyzardKing:
Sounds good.
The important thing for now is to have everything working on 16.04.
The arduino ide should have been the last thing, as far as I can tell

didrocks:
Sorry for the delay, the past month have been crazy (in the last 4 weeks, I've been travelling 3 of them!). Merging it now :)

PR #303, opened 2016-04-22
Remove openjdk dependency for arduino

LyzardKing:
A version of java 8 is bundled in the archive, so it's not needed as a dependency.

This should be the last thing that had dependencies incompatible with xenial


didrocks:
Perfect! Removed as well from the tests dep, looking great!
+1

PR #314, opened 2016-05-12
Fix node_modules folder setting

LyzardKing:
In the nodejs post-install I set the global folder to be a folder in the user directory.
With a recent update to nodejs this was not possible without having the node executable in PATH.
I changed it to manually alter the file .npmrc, if it does not contain already a similar setting.

### DIFF
--- a/umake/frameworks/nodejs.py
+++ b/umake/frameworks/nodejs.py
@@ -78,7 +78,6 @@ class NodejsLang(umake.frameworks.baseinstaller.BaseInstaller):
 
     def post_install(self):
         """Add nodejs necessary env variables and move module folder"""
-        #subprocess.call([os.path.join(self.install_path, "bin", "npm"), "config", "set", "prefix", "~/.node_modules"])
         if not self.prefix_set():
             with open('{}/.npmrc'.format(os.path.expanduser('~')), 'a+') as file:
                 file.write("prefix = ${HOME}/.npm-packages")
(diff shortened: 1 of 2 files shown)


didrocks:
Looking good, the integration tests were failing I guess, right?

Do you mind having a look at the tests to append a check for this file to be created?

LyzardKing:
The problem was that the binary npm calls previously had a relative path, so it worked even before the .profile file was updated. Now it expects to find it in PATH.
It would have been better to use the command we were using previously, because now we have to be careful not to overwrite the file if it contains other settings.
I haven't had much time lately, but it needs tests for:
- no .npmrc
- .npmrc with other settings (not the prefix folder one)
- .npmrc with that settings already there.

I tried on my system manually editing the file and it seems to work properly, but just to be sure we don't remove user-defined settings on a reinstall I'll add some tests.

umake/frameworks/nodejs.py
```diff
@@ -68,9 +68,21 @@ def parse_download_link(self, line, in_download):
...
+            with open('{}/.npmrc'.format(os.path.expanduser('~')), 'r') as file:
+                for line in file.readlines():
+                    if line.startswith("prefix ="):
+                        return True
+        return False
+
     def post_install(self):
         """Add nodejs necessary env variables and move module folder"""
-        subprocess.call([os.path.join(self.install_path, "bin", "npm"), "config", "set", "prefix", "~/.node_modules"])
+        #subprocess.call([os.path.join(self.install_path, "bin", "npm"), "config", "set", "prefix", "~/.node_modules"])
```
didrocks: You did keep this comment, maybe you can clean this out? :)

PR #344, opened 2016-09-17
Fix VSCode binary

### LINKED ISSUE #343: VS Code should be started using the startup script
Currently, VS Code is launched using ```.local/share/umake/ide/visual-studio-code/code```.
It would be better to use the startup script in ```.local/share/umake/ide/visual-studio-code/bin/code```

I think it's just a small modification in [required_files_path](https://github.com/ubuntu/ubuntu-make/blob/16.04/umake/frameworks/ide.py#L726) to use "bin/code"

LyzardKing:
Fix #343


didrocks:
Looks easy enough! Nicely spotted. Merging :)

Issue #335, opened 2016-08-09
.profile not sourced until rebooting


tschf:
I just installed nodejs through umake, and found it wasn't being added to the path after re-launching terminal. This is the output from the install:

```
trent@birroth:~$ umake nodejs
Choose installation path: /home/trent/.local/share/umake/nodejs/nodejs-lang
Downloading and installing requirements                                        |
100% |#########################################################################|
Installing Nodejs Lang
/usr/bin/env: ‘node’: No such file or directory                                |
|##############################################################################|
You need to restart your current shell session for your Nodejs Lang installation to work properly
Installation done
trent@birroth:~$ 
```

I interpreted `restart your current shell` as restarting the terminal (gnome-terminal), but found if I restart the terminal, my PATH is missing the nodejs path that was added to `~/.profile` as part of the installation (and as such, `node` still not on my path). 

Only after I restart my system do I find that the updated `~/.profile` is sourced. Is `shell` referring to my desktop session? Or is it likely I have a bad config somewhere and it should be sourced each time I launch (gnome-)terminal.

didrocks:
`shell` is expected to more understood as a shell session.
Some programs always restart a new shell session when you open a new window. Some forks and reuse the current session like gnome-terminal (from gnome-terminal-server) and so on.
Maybe rephrasing would be a good idea, like "log back"

tschf:
Yeah, sounds good. Or `re-login`? Not sure what's best! 

`You may need to log back in for your NodeJS installation to work properly`?

didrocks:
Sounds good to me! I would accept a good patch changing this terminology + adapting tests ;)

tschf:
Happy to do that. Just reviewing the tests folder (large|medium/test_nodejs.py), can't see exactly what I'd have to adapt, but will let you know if I get stuck. :-)

PR #353, opened 2016-09-29
Update nodejs post install message

tschf:
As discussed in #335 

> Sounds good to me! I would accept a good patch changing this terminology + adapting tests ;)

Not sure what to do for adapting tests. Can you offer any guidance here?
I tried running tests (`./runtests small`, `./runtests medium`) even before any changes (both with steps in README for virtualenv or just installing dependencies on core system) and didn't have success with those. This probably is the right place to discuss those issues, but can provide more info.

Output after the change:

```bash
trent@birroth:~/Projects/ubuntu-make$ ./bin/umake nodejs
Download from https://nodejs.org/dist/latest/SHASUMS256.txt
Choose installation path: /home/trent/.local/share/umake/nodejs/nodejs-lang
Downloading and installing requirements                                                                                                                                                                            |
100% |#############################################################################################################################################################################################################|
Installing Nodejs Lang
|##################################################################################################################################################################################################################|
You may need to log back in for your Nodejs Lang installation to work properly
Installation done
```

cc @didrocks

### DIFF
--- a/umake/frameworks/android.py
+++ b/umake/frameworks/android.py
@@ -136,7 +136,7 @@ class AndroidSDK(umake.frameworks.baseinstaller.BaseInstaller):
         # the SDK manager
         add_env_to_user(self.name, {"PATH": {"value": [os.path.join("$ANDROID_HOME", "tools"),
                                                        os.path.join("$ANDROID_HOME", "platform-tools")]}})
-        UI.delayed_display(DisplayMessage(self.POST_INSTALL_WARN.format(self.name)))
+        UI.delayed_display(DisplayMessage(self.RELOGIN_REQUIRE_MSG.format(self.name)))
 
         # print wiki page message
         UI.delayed_display(DisplayMessage("SDK installed in {}. More information on how to use it on {}".format(
--- a/umake/frameworks/android.py
+++ b/umake/frameworks/android.py
@@ -136,7 +136,7 @@ class AndroidSDK(umake.frameworks.baseinstaller.BaseInstaller):
         # the SDK manager
         add_env_to_user(self.name, {"PATH": {"value": [os.path.join("$ANDROID_HOME", "tools"),
                                                        os.path.join("$ANDROID_HOME", "platform-tools")]}})
-        UI.delayed_display(DisplayMessage(_(super().POST_INSTALL_WARN).format(self.name)))
+        UI.delayed_display(DisplayMessage(self.POST_INSTALL_WARN.format(self.name)))
 
         # print wiki page message
         UI.delayed_display(DisplayMessage("SDK installed in {}. More information on how to use it on {}".format(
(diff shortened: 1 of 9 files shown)


didrocks:
Hey! Thanks for the patch, the changes looks good to me!

After checking anyway, there is no tests that ensure that this is written down (which is ok, as just an user information message).
However, do you mind applying the same changes to other frameworks? scala, go, android, switch, kotlkin, dart & rust are using those. Would be better to have all of them using the same message :)

Thanks in advance!

tschf:
I thought it might not be a bad idea to declare the message once (baseinstaller.py) and reference that from the sub-classes. Or do you prefer to keep them defined in each relevant framework?

Manually tested on:

- android
- dart
- go
- kotlin
- nodejs
- rust
- scala
- swift

umake/frameworks/android.py
```diff
@@ -136,8 +136,7 @@ def post_install(self):
         # the SDK manager
         add_env_to_user(self.name, {"PATH": {"value": [os.path.join("$ANDROID_HOME", "tools"),
                                                        os.path.join("$ANDROID_HOME", "platform-tools")]}})
-        UI.delayed_display(DisplayMessage(_("You need to restart your current shell session for your {} installation "
-                                            "to work properly").format(self.name)))
+        UI.delayed_display(DisplayMessage(_(super().POST_INSTALL_WARN).format(self.name)))
```
didrocks: as it's inherited, you can replace `super()` by `self`, isn't it?

umake/frameworks/baseinstaller.py
```diff
@@ -41,6 +41,11 @@
 class BaseInstaller(umake.frameworks.BaseFramework):
 
     DIRECT_COPY_EXT = ['.svg', '.png', '.ico', '.jpg', '.jpeg']
+    # Framework environment variables are added to `~/.profile` which may
+    # require logging back into your session for the changes to be picked up.
+    # Use `POST_INSTALL_WARN` to alert users to this fact, in `post_install`
+    # function. {} in replaced with the framework name at runtime.
+    POST_INSTALL_WARN = "You may need to log back in for your {} installation to work properly"
```
didrocks: I would change the variable name to `RELOGIN_REQUIRE_MSG`.

Also, this is the variable that needs to have = _("…"). So, you can remove it from the other calls, and only have it once for translation here (the other calls isn't useful, as gettext won't be able to pick up the text assignement properly).

umake/frameworks/android.py
tschf: Ah yeah - I wasn't sure, and had tested with `super()` first so went with that. Will update.

tschf:
Just pushed those requested changes.

didrocks:
Reviewed the changes and everything looks good (and small tests pass!). Good work, thanks again for your contribution there! :+1: 
Merging :)

tschf:
No probs! Thanks for the great tool!

PR #358, opened 2016-10-07
GitHub

LyzardKing:
Adds a check if the download_url is set.


didrocks:
Looking good!
Do you think it's possible to create a medium tests for getting those failing at first?

LyzardKing:
Would that not be a case like the download page changed?
Do you think it needs another test?

didrocks:
Ah, actually, you are probably right, the failure was due to the stacktrace, hence the exit != 0, but we never looked at the output (that it exits cleanly).
So yeah, we will need at some point thinking about enabled checks to ensure we exit cleanly (can be no "Traceback" keyword in the output), even if we expect a non 0 return value.

But ok, good for now, merging! Thanks :)

PR #359, opened 2016-10-18
Initial Eclipse Che commit

LyzardKing:
Working, still needs the tests to be fixed.

### DIFF
--- a/umake/frameworks/ide.py
+++ b/umake/frameworks/ide.py
@@ -955,3 +955,43 @@ class SpringToolsSuite(umake.frameworks.baseinstaller.BaseInstaller):
                                                                             exec='"{}" %f'.format(self.exec_path),
                                                                             comment=_(self.description),
                                                                             categories=categories))
+
+
+class EclipseChe(umake.frameworks.baseinstaller.BaseInstaller):
+
+    def __init__(self, category):
+        super().__init__(name="Eclipse Che", description=_("Eclipse Che is a developer workspace server and cloud IDE"),
+                         category=category, only_on_archs=['amd64'],
+                         download_page='https://www.eclipse.org/che/download/',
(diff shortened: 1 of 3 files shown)


umake/frameworks/ide.py
```diff
@@ -955,3 +955,42 @@ def post_install(self):
...
+    def __init__(self, category):
+        super().__init__(name="Eclipse Che", description=_("Eclipse Che is a developer workspace server and cloud IDE"),
+                         category=category, only_on_archs=['amd64'],
+                         download_page='https://www.eclipse.org/che/download/',
+                         desktop_filename="eclipse-che.desktop",
+                         required_files_path=["bin/che.sh"],
+                         dir_to_decompress_in_tarball="eclipse-che*",
+                         packages_requirements=['openjdk-8-jre', 'maven', 'docker.io'])
+        self.icon_url = 'https://www.eclipse.org/che/images/logo-eclipseche.svg'
+        self.icon_filename = 'eclipse-che.svg'
```
didrocks: You can maybe set that variable as the basename of self.icon_url?

LyzardKing:
I'm still trying to figure out what to put in the launcher.
The ideal thing could be to start che.sh (the server) and xdg-open to localhost:<port>
That is how to actually start the ide. Or we can remove the .desktop and symlink the che.sh
I actually have no idea on how it works.
@rosolko Do you have any suggestion?

rosolko:
@LyzardKing You are correct about server.
After installation we need to perform some action:
`chmod +x /usr/local/bin/che`
And after that we can start che.sh
After that moves we can go to localhost and use IDE.
Desktop and symlink need to start server before using.

LyzardKing:
The only version of eclipse che now available is the docker image. So this installer is no longer needed.

PR #360, opened 2016-10-19
Android Studio duplicate icons in launcher fix

### LINKED ISSUE #270: Android Studio Unity launcher shows duplicate icons (Matching fails)
If I create desktop file for android studio from within android studio everything works great i.e. there is only one icon for android studio in the launcher, but if the icon is created by Ubuntu Make then there are two launchers shown in the launcher after android studio is started.

I compared the two desktop files and apparently the one that is created by android studio contains the line:
  StartupWMClass=jetbrains-studio

while the once created by umake contains:
  StartupWMClass=jetbrains-android-studio

MikeChou:
fixes #270


LyzardKing:
Thanks. Sorry it took a while ;)

PR #370, opened 2016-11-10
Update tr.po

WhiteSymmetry:


didrocks:
Looking good! Merging as well :) Thanks

PR #373, opened 2016-11-12
Removed jayatana dependency from android-studio

LyzardKing:
Android studio is built on jetbrains idea, and includes a version of openjdk-jre.
Jayatana in this case does nothing.


didrocks:
:+1:

PR #387, opened 2016-12-02
android: add a few fall-back variables that might be used by some tools

sschuberth:
In particular, ANDROID_NDK_HOME is used by Android Studio 2.2.


didrocks:
This looks good to me! Thanks for adding them with the detailed informations :)
Merging

PR #412, opened 2017-02-22
Return the pexpect reply so we can use multiple option expect.

LyzardKing:
It was already used in the eap tests, but I guess it was never actually running.
@didrocks, could this break anything else running in the tests?

### DIFF
--- a/tests/large/__init__.py
+++ b/tests/large/__init__.py
@@ -143,15 +143,13 @@ class LargeFrameworkTests(LoggedTestCase):
             else:  # nothing found:
                 raise BaseException("We didn't find an expected WARNING or ERROR in {}".format(content))
 
-    def expect_and_no_warn(self, expect_query, timeout=-1, expect_warn=False):
-        """run the expect query and check that there is no warning or error
-
-        It doesn't fail on the given timeout if stdout is progressing"""
+    def return_and_wait_expect(self, expect_query, timeout=-1):
+        """run the expect query and return the result"""
         output = ""
         continue_expect = True
         while continue_expect:
             try:
-                self.child.expect(expect_query, timeout=timeout)
+                result = self.child.expect(expect_query, timeout=timeout)
                 continue_expect = False
             except pexpect.TIMEOUT:
                 # stalled during timeout period
@@ -163,6 +161,13 @@ class LargeFrameworkTests(LoggedTestCase):
             except:
                 print(self.child.before)
                 raise
+        return result
+
+    def expect_and_no_warn(self, expect_query, timeout=-1, expect_warn=False):
+        """run the expect query and check that there is no warning or error
+
+        It doesn't fail on the given timeout if stdout is progressing"""
+        self.return_and_wait_expect(expect_query, timeout, expect_warn)
         self.assert_for_warn(self.child.before, expect_warn)
 
     def wait_and_no_warn(self, expect_warn=False):


didrocks:
No, it can't, as you are just returning something which doesn't affect the caller (and there is no magic in python for a method being sync turning async if the return value is read).

I wonder though if we shouldn't move that to 2 functions. As it sounds weird to have "expect_and_no_warn" returning something, no?
Could we have another method with the content and returning result and expect_and_no_warn() calling that one (ignoring returned value)? In that way, for EAP, you will just call the first method?

LyzardKing:
I thought it was fine there since it's not a warning, just a status update...
But sure. I'll change that now

LyzardKing:
The new method would have all the content from the current one, including the return. So what would the other one do? I don't see the use for another method..

didrocks:
What I see is:
```def return_and_wait_pexpect(self, expect_query, timeout=-1):
<contain everything in expect_and_no_warn() but the last self.assert_for_warn(self.child.before, expect_warn)>
return result

Then:
def expect_and_no_warn(self, expect_query, timeout=-1, expect_warn=False):
  result =  return_and_wait_pexpect(expect_query, timeout=-1)
  self.assert_for_warn(self.child.before, expect_warn)
```

That way, current method does the same, and the EAP method will use ` return_and_wait_pexpect` to get the returned result (+ optionally call as well self.assert_for_warn(self.child.before, expect_warn))

Making more sense?

LyzardKing:
Ah ok. So in one of them (the eap one, although it might be used by other things) we don't call directly  the assert_for_warn.

LyzardKing:
Is this what you meant?

tests/large/__init__.py
```diff
@@ -143,15 +143,13 @@ def assert_for_warn(self, content, expect_warn=False):
...
-
-        It doesn't fail on the given timeout if stdout is progressing"""
+    def return_and_wait_expect(self, expect_query, timeout=-1, expect_warn=False):
+        """run the expect query and return the result"""
         output = ""
         continue_expect = True
         while continue_expect:
-            try:
-                self.child.expect(expect_query, timeout=timeout)
+            try:
```
didrocks: Mind removing those extra spaces?

tests/large/__init__.py
```diff
@@ -163,7 +161,15 @@ def expect_and_no_warn(self, expect_query, timeout=-1, expect_warn=False):
...
                 raise
+        return result
+
+    def expect_and_no_warn(self, expect_query, timeout=-1, expect_warn=False):
+        """run the expect query and check that there is no warning or error
+
+        It doesn't fail on the given timeout if stdout is progressing"""
+        self.return_and_wait_expect(expect_query, timeout=-1, expect_warn=False)
         self.assert_for_warn(self.child.before, expect_warn)
+
```
didrocks: Remove that line (only one line between methods)

tests/large/__init__.py
```diff
@@ -163,7 +161,15 @@ def expect_and_no_warn(self, expect_query, timeout=-1, expect_warn=False):
             except:
                 print(self.child.before)
                 raise
+        return result
+
+    def expect_and_no_warn(self, expect_query, timeout=-1, expect_warn=False):
+        """run the expect query and check that there is no warning or error
+
+        It doesn't fail on the given timeout if stdout is progressing"""
+        self.return_and_wait_expect(expect_query, timeout=-1, expect_warn=False)
```
didrocks: Don't copy the default values here, keep timeout.

tests/large/__init__.py
```diff
@@ -143,15 +143,13 @@ def assert_for_warn(self, content, expect_warn=False):
             else:  # nothing found:
                 raise BaseException("We didn't find an expected WARNING or ERROR in {}".format(content))
 
-    def expect_and_no_warn(self, expect_query, timeout=-1, expect_warn=False):
-        """run the expect query and check that there is no warning or error
-
-        It doesn't fail on the given timeout if stdout is progressing"""
+    def return_and_wait_expect(self, expect_query, timeout=-1, expect_warn=False):
```
didrocks: Reread what I wrote, no expect_warn here as it's not used in that method :)

tests/large/__init__.py
LyzardKing: Sorry...a quick copy-paste

LyzardKing:
Should be fine now

didrocks:
Perfect! Looks good to me. Thanks :)
:+1:

LyzardKing:
Thank you!

PR #426, opened 2017-03-09
Error warning

LyzardKing:
Adds the error warning to update umake if an install fails and there is an update on github.
The tests now work, for failed downloads and: a new version of umake, no new umake version, the umake github page missing (404 error)


didrocks:
There are few changes I want to make, but that could be in the next PR. I'm going to add our test as well. Nice work! :)

PR #428, opened 2017-03-10
Atom apm

### LINKED ISSUE #411: Atom install - apm missing from path
I think apm is a common command tool to use alongside atom, but notice it doesn't come in the path when installing through umake. 

```
trent@birroth:~$ type -a atom
atom is /home/trent/.local/share/umake/bin/atom
atom is /usr/bin/atom
trent@birroth:~$ type -a apm
apm is /usr/bin/apm
```

Looking at the install files from the deb installation:

```
trent@birroth:~$ dpkg -L atom | grep apm$
/usr/share/atom/resources/app/apm
/usr/share/atom/resources/app/apm/bin/apm
/usr/share/atom/resources/app/apm/node_modules/.bin/apm
/usr/bin/apm
```

Thus it looks like possibly a binary should be set up for `~/.local/share/umake/bin/apm` to point to `~/.local/share/umake/ide/atom/resources/app/apm/node_modules/.bin/apm`

..

edit. Just trying to dig around to see how this is done, looks like the link in the path (`~/.local/share/umake/bin`) is set up through baseinstaller.py:decompress_and_install_done, by way of exec_link_name and exec_path (single), so possibly not currently set up to support multiple binaries being added into `umake/bin` (initially, I was thinking the `required_files_path`array was set up so a list of binaries you wanted to be linked could be referenced, but I think this is just used to check when uninstalling - possibly some confusion on my part here)

LyzardKing:
I added the atom apm binary to the bin folder as a symlink.
Closes #411


didrocks:
Looks good to me! Let's merge this as well :) Thanks!

PR #429, opened 2017-03-10
Fix Unity download to use DownloadCenter

LyzardKing:
@didrocks Just going to randomly leave this here...
The medium test works. If something's not right we can still just ignore it for now


LyzardKing:
@didrocks This was closed by mistake..

LyzardKing:
The large and medium tests work.
Merging to master after the travis test.

PR #430, opened 2017-03-10
Add spring tools

LyzardKing:
Rebuilt copying the eclipse custom downloader.
The large test works. The medium test fails, with the error:
umake ide: error: argument framework: invalid choice: 'spring-tools-suite'


LyzardKing:
@didrocks Any idea why the mediumt test fails?

didrocks:
This is typical that you don't have the correct packages available in the docker container. So, packages are not available to be installed -> framework not available to the user on this arch/configuration.

You can start the container by hand and inspect its content via apt-cache policy for instance and looks at required packages.

LyzardKing:
you're right. They were looking for a mock repo called eclipse, but for eclipse we use android.
So now all the tests work.

didrocks:
\o/

didrocks:
That looks good to me! Let's merge it :)

PR #435, opened 2017-03-22
Make nodejs read the download page instead of fixed shasum files

### LINKED ISSUE #434: nodejs LTS version old
Fresh install of Ubuntu 16.04 LTS. Installed umake version 17.03. Then installed nodejs with --lts. nodejs version installed is v4.8.1. I would expect at least v6.9.1.

How would I go about updating this through umake?

LyzardKing:
Fix #434 
@didrocks  I'm fixing the tests. Do you think this is a good way of doing things?
I parse the download page for the shasum file we were using earlier and then overwrite the self.download_file with that. This should minimize the difference from baseInstaller.


didrocks:
That looks good to me and makes totally sense!
Your approach indeed is more inline with what we did in other frameworks.

I guess there was no stable pointer like `https://nodejs.org/dist/latest-lts/SHASUMS256.txt`, correct?

LyzardKing:
Not that I could find. If it exists we can revert this change.

didrocks:
Ok then, let's merge it now! May thanks :)

PR #436, opened 2017-03-24
Add phantomjs to web framework

LyzardKing:
This adds phantomjs downloaded from bitbucket (via https).
The medium and large tests work


tests/large/test_web.py
```diff
@@ -115,6 +116,50 @@ def language_file_exists(self, language):
         return self.path_exists(os.path.join(self.installed_path, "dictionaries", "{}.aff".format(language)))
 
 
+class PhantomJSTests(LargeFrameworkTests):
+    """The default Go google compiler."""
```
didrocks: It's not the Go compiler :)

tests/large/test_web.py
```diff
@@ -115,6 +116,50 @@ def language_file_exists(self, language):
...
+        self.child.sendline("")
+        self.expect_and_no_warn("Installation done", timeout=self.TIMEOUT_INSTALL_PROGRESS)
+        self.wait_and_close()
+
+        self.assert_exec_exists()
+        self.assertTrue(self.is_in_path(self.exec_path))
+
+        # compile a small project
+        output = subprocess.check_output(self.command_as_list(compile_command)).decode()\
+            .replace('\r', '').replace('\n', '')
```
didrocks: You can escape the \r\n by just slicing your output if you know they will be present. Or comparing with the assertEqual having \r\n.

tests/large/test_web.py
LyzardKing: I wonder how that got there...

tests/large/test_web.py
LyzardKing: It works if I instead use [:-1]
Should we then change it also in the go test (I have to admit I copied that...)

didrocks: please, feel free to change wherever it makes sense :)
Just ensure this work both in a large tests and within the medium tests, as we don't have the exact same configuration there.

LyzardKing: Sure ;) testing now for phantomjs in medium test

LyzardKing: it works if I replace the mock command from the echo to:
`printf "hello, world\r"`
Is that ok?
For now I'll just fix it on phantom. then I'll do another pr to fix other tests later on.

didrocks: Working for both medium and large tests? If so, yeah, feel free :)

LyzardKing: And done. Should be good now.

didrocks:
Merging then! Thanks a lot :)

PR #320, opened 2016-06-09
Add insiders build for vscode

overbid:
No test, somebody help me pls.

---

This change is [Reviewable](https://reviewable.io/reviews/ubuntu/ubuntu-make/320)

PR #437, opened 2017-03-27
Add support for VSCode Insiders build

LyzardKing:
Merge #320
@didrocks The large test is working.
The medium test fails, it can't find the process to stop.
In the container the process seems to be working properly as far as I can tell.


tests/medium/__init__.py
```diff
@@ -135,7 +135,7 @@ def check_and_kill_process(self, process_grep, wait_before=0, send_sigkill=False
                                                                                       "check_and_kill_process"),
                                                                          send_sigkill,
                                                                          " ".join(process_grep))))[0]:
-            raise BaseException("The process we try to find and kill can't be found".format(process_grep))
+            raise BaseException("The process we try to find and kill can't be found: {}".format(process_grep))
```
LyzardKing: Was this omitted for a reason?

didrocks: No, it wasn't :)

LyzardKing:
in the install folder bin/code calls code.
So we'd need to close code instead of bin/code.
At least, this is what works with the large tests.
I can't get the medium tests to work now...

didrocks:
As discussed by emails, the medium tests actually works! Merging this thus :) Thanks!

PR #438, opened 2017-03-29
Fix ide medium tests

LyzardKing:
This pr simply moves the mock jetbrains archives to the correct location and
adds the server download.jetbrains.com to the docker image.
It also fixes the netbeans medium tests, that had a mix of two different versions and was not working


didrocks:
Nicely spotted and well-done! :+1: for merging

PR #449, opened 2017-04-19
Add force default location

LyzardKing:
I added a --force option to automatically set the default
destination folder in which to install the framework.
@didrocks Let me know if this is something that needs tests or other checks.
@TomaszDom If you want to test it, let me know if it's what you intended.


LyzardKing:
I'm not sure why the tests are failing.
I'll revisit this when I have more time.

TomaszDom:
Will this feature be revisited?

LyzardKing:
This feature needs testers and some help figuring out the issues.
There are some tests that fail, and we need to figure out why.
Then we can check if it works properly with all configurations
and if all is well we can merge. But only then, since it could
introduce new issues while solving this one.

PR #442, opened 2017-04-03
Properly remove vscode from web framework

LyzardKing:
> WARNING: Visual Studio Code is now in the ide category, please refer it from this category from now on. This compatibility will be dropped after Ubuntu 16.04 LTS.

Should we completely remove visual-studio-code from the web framework now?

Issue #323, opened 2016-06-12
Firefox Developer Edition: 1. Too big Icon 2. visible under normal Firefox icon when opened


jnachtigall:
![anim](https://cloud.githubusercontent.com/assets/3307404/15989346/70d114fe-3074-11e6-9a4f-97c7a536f60b.gif)

Two small nitpicks here:

**One:** At the beginning of the screencast: **Firefox Developer Edition icon is too big**. This is a UX inconsistency. Steps to reproduce:
1. Search for application "Firefox". 
2. Notice the difference in size between normal and developer editon. Icon of developer is way too big

**Two:** Later in the screencast I open the Developer Edition: When opening Firefox developer Editon, then the **edition window is attached to the normal firefox icon**. This is very irritating, especially when I run normal Firefox and Editon at the same time because I cannot easily differentiate the open windows.

abdulhannanali:
@jnachtigall The second issue about the Firefox Developer Edition not being blue is not the issue specific to Ubuntu Make. I suppose if the first one too is, I just did a manual install of Firefox Developer Edition and even If I don't have the normal Firefox open, it gives the same icon the one with the orange in it. However, the internals are totally different.

LyzardKing:
The second issue is solved.
The first one is an issue to raise with Mozilla.
At the moment I'll close this issue, since there's nothing regarding umake.

PR #463, opened 2017-07-17
Change the launchers to use the symlink

### LINKED ISSUE #448: How to get Unity Launcher for Visual Studio Code and Atom to respect .bashrc
I am using Ubuntu 16.04 in virtualbox. There is a bug in electron that causes visual studio code and atom to open with a blank screen with the gpu enabled. To stop this, instead of launching

`visual-studio-code` and `atom`

I need to launch the application as `visual-studio-code --disable-gpu` and `atom --disable-gpu`.

When I add aliases to the .bashrc file in my home directory,
~~~~~~~~~~
# Aliases for Visual Studio Code and Atom to disable gpu upon launch
alias visual-studio-code='visual-studio-code --disable-gpu'
alias atom='atom --disable-gpu'
~~~~~~~~~~

Visual Studio Code and Atom launch fine from the terminal. However, When I click the app icon, it still launches without disabling the gpu for the application. How do I address this?

LyzardKing:
Change the launchers to use the symlink instead of the full path
Add the full path as TryExec (mainly for the tests)
Fixes #448

### DIFF
--- a/umake/tools.py
+++ b/umake/tools.py
@@ -347,7 +347,7 @@ def add_exec_link(exec_path, destination_name):
     os.symlink(exec_path, full_dest_path)
 
 
-def get_application_desktop_file(name="", icon_path="", exec="", comment="", categories="", extra=""):
+def get_application_desktop_file(name="", icon_path="", try_exec="", exec="", comment="", categories="", extra=""):
     """Get a desktop file string content"""
     return dedent("""\
                 [Desktop Entry]
@@ -355,12 +355,13 @@ def get_application_desktop_file(name="", icon_path="", exec="", comment="", cat
                 Type=Application
                 Name={name}
                 Icon={icon_path}
+                TryExec={try_exec}
                 Exec={exec}
                 Comment={comment}
                 Categories={categories}
                 Terminal=false
                 {extra}
-                """).format(name=name, icon_path=icon_path, exec=exec,
+                """).format(name=name, icon_path=icon_path, try_exec=try_exec, exec=exec,
                             comment=comment, categories=categories, extra=extra)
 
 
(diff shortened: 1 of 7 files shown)


LyzardKing:
The second commit (Firefox) would solve #323

LyzardKing:
@didrocks What do you think of this proposal?
It could fix a couple of issues

tests/medium/test_web.py
```diff
@@ -65,17 +65,3 @@ def test_install_with_changed_download_page(self):
         umake_command = self.command('{} web phantomjs'.format(UMAKE))
         self.bad_download_page_test(umake_command, download_page_file_path)
         self.assertFalse(self.path_exists(self.exec_path))
-
-
-class VisualStudioCodeInContainer(ContainerTests, test_web.VisualStudioCodeTest):
```
didrocks: That should probably be in another PR? (the finale migation of vsc in IDE category)

umake/tools.py
```diff
@@ -447,13 +448,16 @@ def add_env_to_user(framework_tag, env_dict):
...
+        profile_filepath = profile_filepath.replace("profile", "bashrc")
+
     with open(profile_filepath, "a", encoding='utf-8') as f:
         f.write(profile_tag.format(framework_tag))
         for env in envs_to_insert:
             value = envs_to_insert[env]
             logger.debug("Adding {} to user's {} for {}".format(value, env, framework_tag))
             export = ""
-            if env != "PATH":
+            if env != "PATH" and not env.startswith("alias "):
```
didrocks: Should we have a test for this case?

tests/medium/test_web.py
LyzardKing: Yes..this was part of #442
It's fixed in my local branch...but it's not uploaded yet

umake/tools.py
LyzardKing: We have now a simple test in test_tools.py

LyzardKing:
Moved the alias setting in tools to the new firefox PR.

Issue #459, opened 2017-06-23
tests/small/test_decompressor.py tests a


mwhudson:
As of 229e4ee8dbb5557935eb7c4a7ed490c6c5267512:

```
ubuntu@artful:~/ubuntu-make$ python3.6 `which nosetests3` tests/small/test_decompressor.py 
...........FFFF
======================================================================
FAIL: We decompress a valid zip file successfully
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ubuntu/ubuntu-make/tests/small/test_decompressor.py", line 130, in test_decompress_zip
    self.assertIsNone(results[fd].error)
AssertionError: "'str' object has no attribute 'external_attr'" is not None
-------------------- >> begin captured logging << --------------------
umake.decompressor: INFO: Requesting decompression to /tmp/tmplv6a5hle
umake.decompressor: DEBUG: Extracting to /tmp/tmplv6a5hle
umake.decompressor: DEBUG: zip file
umake.decompressor: ERROR: A decompression to /tmp/tmplv6a5hle failed: 'str' object has no attribute 'external_attr'
Traceback (most recent call last):
  File "/usr/lib/python3.6/concurrent/futures/thread.py", line 55, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/ubuntu/ubuntu-make/umake/decompressor.py", line 105, in _decompress
    archive.extractall(tempdest)
  File "/usr/lib/python3.6/zipfile.py", line 1501, in extractall
    self._extract_member(zipinfo, path, pwd)
  File "/home/ubuntu/ubuntu-make/umake/decompressor.py", line 47, in _extract_member
    mode = member.external_attr >> 16 & 0x1FF
AttributeError: 'str' object has no attribute 'external_attr'
umake.decompressor: INFO: Decompression to /tmp/tmplv6a5hle finished
umake.decompressor: INFO: All pending decompression done to ['/tmp/tmplv6a5hle'] done.
--------------------- >> end captured logging << ---------------------

======================================================================
FAIL: We decompress a valid zip file successfully
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ubuntu/ubuntu-make/tests/small/test_decompressor.py", line 49, in tearDown
    super().tearDown()
  File "/home/ubuntu/ubuntu-make/tests/tools/__init__.py", line 64, in tearDown
    self.assertEqual(self.error_warn_logs.getvalue(), "")
[...]

PR #466, opened 2017-07-25
Fix ZipFileWithPerm compatability with internal 3.6 API

xnox:
Internal implementation of the _extract_member private API has changed in 3.6. Adapt the override function to be compatible with 3.6 and pre-3.6.

Fixes https://github.com/ubuntu/ubuntu-make/issues/459


didrocks:
Looks good and tests still pass! Thanks a lot @xnox :)

PR #478, opened 2017-10-13
Change preferred install method to snap in README

LyzardKing:
Edit the Readme to indicate that the snap is the currently preferred install method.
If there are no further suggestions I'll merge this later today.


didrocks:
+1 from my side

PR #479, opened 2017-10-13
Version

LyzardKing:
Not to merge yet.
Just to discuss the issue, I think we could add --version and --upgrade options.
version would parse the download page and return the latest available version.
upgrade would compare that with the installed version and if a condition is met (remote version > local)
it could reinstall.
The upgrade option could be explicitly enabled for frameworks that do not auto upgrade:
Not enabled for example for Intellij ides, firefox,
Enabled for atom, vscode, go, and others...
At the moment there are a couple of tests(one is not working yet), and it's only properly activated for atom.


LyzardKing:
Replaced by #506

PR #503, opened 2017-12-14
Fix environment variables for Android SDK and NDK.

titobrasolin:
The add_env_to_user function must be called only once, it takes a dictionary of environment variables as a second parameter.


LyzardKing:
I'm just testing it now. If it works I'll merge.
I would be nice to add a test to verify an incorrect insertion...but we'll see

LyzardKing:
Thank you @titobrasolin!

PR #506, opened 2017-12-18
Version update

### LINKED ISSUE #74: Support updating tools
Ok, the Twitter website just crashed my Firefox while I was in the middle of writing this up, so this writeup is gonna be shorter. :p

Some tools can update themselves - stuff from PPAs, JetBrains, Eclipse, Android stuff I think. This doesn't concern these tools.

Some tools can't, like go-lang and probably any other language support we add in the future. We should have an upgrade workflow for these tools.

Actually upgrading the tools should be easy due to our policy of always getting the latest upstream - remove&install. What we should improve is the way users know what needs to be updated.

A subset of our tools should support reporting the version of a tool back. So when we install a tool, the version goes into the config file.

Then users can ask umake - list all tools that have updates. We go through all installed tools from the config file, check if their modules support versioning. If they do, fetch latest upstream version, compare with the current version (if current version is missing, because the tool was installed before we added support for versions, flag as 'update available') and list everything that can be updated.

LyzardKing:
Add --version and --update commands for supported frameworks.
This simply compares the installed version with the upstream available version.

The auto-updating frameworks would not be supported.
As an example, firefox-dev only returns the version.
Atom, Sublime-text and go are updatable

Fixes #74

### DIFF
--- a/tests/data/testframeworks/baseinstallerfake.py
+++ b/tests/data/testframeworks/baseinstallerfake.py
@@ -82,12 +82,18 @@ class BaseFramework(umake.frameworks.baseinstaller.BaseInstaller):
                          checksum_type=ChecksumType.sha1,
                          dir_to_decompress_in_tarball="base-framework-*",
                          desktop_filename="base-framework.desktop",
-                         required_files_path=[os.path.join("bin", "studio.sh")], **kwargs)
+                         required_files_path=[os.path.join("bin", "studio.sh")],
+                         updatable=True, **kwargs)
 
         arch = platform.machine()
         self.tag = 'id="linux-bundle64"'
         if arch == 'i686':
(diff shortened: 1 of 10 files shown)


Xerus2000:
Guys I really want this to be in umake! Why is it still not merged and not even adressed?

LyzardKing:
@Xerus2000 There has been ongoing work..
But as specified in the related issues, it's not something easy to implement.
It's updated, rebased on master...but it'll not be merged for the moment.

LyzardKing:
The main issue is recreating the correct cli instrucions for an already installed framework.
We do not have access to the original arguments.

Xerus2000:
well you could write them into the config file too, that seems to me like the easiest option

Tonymmm (approved):

PR #511, opened 2017-12-27
Add docker to travis

LyzardKing:
Having the medium tests run frequently helps to avoid making a change and not checking the appropriate tests.
This PR introduces the docker image on travis ci, so we can run the medium tests as well as the small ones.
I'm using this to correct all standing issues in the testing interface, and checking if some frameworks have been needing support

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -14,3 +14,6 @@ coverage.xml
 *,cover
 *.log
 .git
+**/__pycache__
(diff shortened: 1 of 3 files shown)


LyzardKing:
I'm not sure why, but the third time I ran the test it succeeded.
The first two I got an error with the webstsorm eap install...

Issue #509, opened 2017-12-27
Add support to Godot Game Engine


mstuttgart:
It would be nice if support for Godot Game Engine. :smile: 

* https://godotengine.org/
* https://godotengine.org/download

This engine is opensource and free. Please consider it.

I can add support and send a PR, but I not find `ubuntu-make` docs.

LyzardKing:
Hi! If you could start a Pr it would be great.
You can have a look at the games.py file, you add a class there.
As an example to copy you can look at the SublimeText class in ide.py (that is the simplest, and it seems to be the use case here).
After that is set up You need to add entries in tests (both medium and large.) I'll help with that!
Let me know if you get it running!

mstuttgart:
Thanks @LyzardKing. I'm going to start working on it.

LyzardKing:
Let me know if you need any help

PR #512, opened 2017-12-29
[NEW] Add godot engine support

mstuttgart:
* Add Godot Game Engine support
* Add tests

Tested in: Linux Mint 18.3 - x64

close https://github.com/Ubuntu/ubuntu-make/issues/509


umake/frameworks/games.py
```diff
@@ -25,9 +25,11 @@
 from gettext import gettext as _
 import logging
 import os
+from os.path import join
```
LyzardKing: This is not needed since you used the full 'os.path.join', so it can be removed

umake/frameworks/games.py
```diff
@@ -385,3 +387,70 @@ def post_install(self):
...
+    def parse_download_link(self, line, in_download):
+        """Parse Godot download links"""
+        url = None
+        if '.zip' in line:
+            p = re.search(r'href="([^<]*{}.zip)"'.format(self.arch_trans[get_current_arch()]), line)
+            with suppress(AttributeError):
+                url = p.group(1)
+        return ((url, None), in_download)
+
+    def decompress_and_install(self, fds):
```
LyzardKing: You don't need to do this, the other frameworks only override the parts they need (the url selection mostly)

umake/frameworks/games.py
```diff
@@ -385,3 +387,70 @@ def post_install(self):
...
+        """Get Godot executable name from url"""
+        super().decompress_and_install(fds)
+        exec_url = self.download_requests[0].url
+        self.exec_name = os.path.splitext(os.path.basename(exec_url))[0]
+
+    def save_icon(self, download_result):
```
LyzardKing: This as well..you only need to set the icon in the post_install part

umake/frameworks/games.py
```diff
@@ -385,3 +387,70 @@ def post_install(self):
...
+                         download_page="https://godotengine.org/download",
+                         desktop_filename="godot.desktop",
+                         required_files_path=["godot"],
+                         dir_to_decompress_in_tarball="",
+                         icon_filename='icon.png',
+                         **kwargs)
+        self.icon_url = os.path.join("https://raw.githubusercontent.com/godotengine/godot/",
+                                     "master",
+                                     self.icon_filename)
+        self.exec_name = ''
```
LyzardKing: exec_name is set automagically, don't override it (it's set as the first element in the required_files_path)

umake/frameworks/games.py
mstuttgart: If a remove, its cause `import error`

umake/frameworks/games.py
mstuttgart: The godot executable name is `Godot_v2.1.4-stable_x11.64`. I need to override this method to get the executable name from url.

umake/frameworks/games.py
mstuttgart: The zip file from godot don't has icon file. I need to download from other source.

umake/frameworks/games.py
mstuttgart: Executable name again. I need to this name to rename Godot executable `Godot_v2.1.4-stable_x11.64` after install.

umake/frameworks/games.py
mstuttgart: :smile: my mistake

LyzardKing:
I can open a pr on your

mstuttgart:
Sorry @LyzardKing.

How I download the icon file? The downloaded .zip file from https://godotengine.org/download not has icon file inside it. How I pass the correct image url to `post_install`?

umake/frameworks/games.py
```diff
@@ -385,3 +386,70 @@ def post_install(self):
...
+
+        # Rename Godot executable
+        os.rename(os.path.join(self.install_path, self.exec_name),
+                  os.path.join(self.install_path, self.exec_link_name))
+
+        # Get icon path
+        icon_path = join(self.install_path, self.icon_filename)
+
+        comment = self.description
+        categories = "Development;IDE;"
```
LyzardKing: Just add self.description and the categories in the create launcher in the following lines.
Also, remove the empty lines

umake/frameworks/games.py
LyzardKing: This is good...

umake/frameworks/games.py
LyzardKing: Remove this part and as I said keep everything as is. The link will simply be "godot"

LyzardKing:
Sorry, the way you fetch the icon is good. Change the exec_name stuff and the  decompress_and_install method then it should be fine.
After that we need to set up the medium test

LyzardKing:
I just realized that godot releases as an appimage...
I'm not sure what we want/need to do to set this up properly...

LyzardKing:
Umake and appimages, to some extent, solve the same issue (an easy way to install things).
I worry that putting the two together could cause issues (They both have desktop launchers, appimages have a deamon to automatically set them up in the desktop, ...)

mstuttgart:
It work fine to me. I guess that appimage not is a problem.

I made the changes. I remove the `decompress_and_install` method, but this made the installer not work so.
A take a screen shoot of AppImage downloaded.

screenshot from 2017-12-30 03-10-57

The .desktop file

``` 
[Desktop Entry]
Version=1.0
Type=Application
Name=Godot
Icon=/home/michell/.local/share/umake/games/godot/icon.png
TryExec=/home/michell/.local/share/umake/games/godot/Godot_v2.1.4-stable_x11.64
Exec=godot
Comment=Multi-platform 2D and 3D opensource game engine
Categories=Development;IDE;
Terminal=false
```
`TryExec` parameter not match.

Issue #515, opened 2018-01-06
Add support for Ren'Py


Dekedro:
Hi, it would be great to have support for Ren'Py - visual novel engine. It's open source and there are archives to download in their page:

https://www.renpy.org/
https://www.renpy.org/latest.html

I'm not familiar with python but I can try to make PR

LyzardKing:
Hi! It could fit in the games.py file.
If you want to have a look at it, it should be fairly easy.
You can look at the Stencyl class in that file.
You can then replace the name, url (use: https://www.renpy.org/latest.html), requirements, and the regex to find  the tarfile.
If you need help let me know

Dekedro:
After some consideration, decided to use renpy from official ubuntu packages: https://packages.ubuntu.com/source/xenial/renpy. I think that it is not efficient to duplicate work.

PR #516, opened 2018-01-07
Support for Ren'Py

Dekedro:
- Added support for Ren'Py
- Added tests

Works on: Ubuntu 16.04.03 x64, development version of Ubuntu 18.04 x64
Issue: #515 
Basically just copy-pasted code from other engines, because I'm not too familiar with codebase of this project and python itself. Detailed review is required.

### DIFF
--- a/tests/medium/test_games.py
+++ b/tests/medium/test_games.py
@@ -53,6 +53,20 @@ class BlenderInContainer(ContainerTests, test_games.BlenderTests):
         self.installed_path = os.path.join(self.install_base_path, "games", "blender")
 
 
+class RenPyInContainer(ContainerTests, test_games.RenPyTests):
+    """This will test the Ren'Py editor inside a container"""
+
+    TIMEOUT_START = 20
+    TIMEOUT_STOP = 10
+
+    def setUp(self):
+        self.hosts = {443: ["www.renpy.org"]}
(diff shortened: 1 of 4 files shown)


tests/medium/test_games.py
```diff
@@ -53,6 +53,20 @@ def setUp(self):
...
 
+class RenPyInContainer(ContainerTests, test_games.RenPyTests):
+    """This will test the Ren'Py editor inside a container"""
+
+    TIMEOUT_START = 20
+    TIMEOUT_STOP = 10
+
+    def setUp(self):
+        self.hosts = {443: ["www.renpy.org"]}
+        self.apt_repo_override_path = os.path.join(self.APT_FAKE_REPO_PATH, 'renpy')
```
LyzardKing: This is not needed since renpy apparently doesn't have other dependencies than need to be installed (is that correct?)

umake/frameworks/ide.py
```diff
@@ -799,7 +799,7 @@ def __init__(self, **kwargs):
                          **kwargs)
 
     def parse_license(self, line, license_txt, in_license):
-        """Parse Android Studio download page for license"""
+        """Parse Visual Studio Code download page for license"""
```
LyzardKing: I'll change this directly, let's keep the PR only relevant to renpy.

PR #548, opened 2018-05-18
Fix Arduino framework and tests (and remove bs4 dependency)

LyzardKing:
This removes the dependency on bs4, 
changing the arduino ide to be closer to the other frameworks.
EDIT: I'm checking now the main tests for arduino, that still need fixing.


didrocks:
If I'm correct, we won't have BeautifulSoup used anymore in ubuntu-make once this is merged, correct?
Can you remove it from debian/control then? Thanks!

LyzardKing:
I did modify the dependencies in Debian/control...did I forget to push? I removed the python-bs4 lines

didrocks:
I missed it! Excellent, thanks, let's merge it :)

PR #549, opened 2018-05-19
Fix Rust and Swift frameworks

LyzardKing:
Swift needs an older version of libicu, not yet compatible with bionic
(At the moment it'll install, but the user needs to find a way to fix this...is there anything we can do?)

Rust needs a change to the symlink in post_install


didrocks:
I would say let's not offer it as an install option until the compatible libicu version is avaible, wdyt?

LyzardKing:
Yes, or... Does livicu have specific numbered versions? So we can set the
explicit versions supported...like we did for openjdk

On May 22, 2018 9:30 AM, "Didier Roche" <notifications@github.com> wrote:

I would say let's not offer it as an install option until the compatible
libicu version is avaible, wdyt?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/ubuntu/ubuntu-make/pull/549#issuecomment-390891087>,
or mute
the thread

didrocks:
As you know more the context than I do, I let you depending on the correct ABI version if it's based on a new ABI.

LyzardKing:
As far as I can tell...I'll test that later today and set the explicit
versions if they're available

LyzardKing:
Done. added the two supported versions of libicu.
We can revisit this when libicu60 is supported.
I changed the packages in the docker test to enable the medium test

didrocks:
Last changes LGTM, feel free to merge! Nice work :)

PR #553, opened 2018-05-21
Try Add DBeaver

patsevanton:
But test output

 ./runtests
  File "./runtests", line 337, in <module>
    if cmd[0] not in ["local", "vm", "remote"]:
IndexError: list index out of range


tests/medium/test_ide.py
```diff
@@ -479,6 +479,20 @@ def setUp(self):
...
 
 
+class DBeaverInContainer(ContainerTests, test_ide.DBeaverTests):
+    """This will test the DBeaver integration inside a container"""
+
+    TIMEOUT_START = 20
+    TIMEOUT_STOP = 10
+
+    def setUp(self):
+        self.hosts = {443: ["dbeaver.io"]}
```
LyzardKing: since you'll be using the github website only you need to put github.com here

umake/frameworks/ide.py
```diff
@@ -857,6 +857,41 @@ def post_install(self):
                         categories="Development;TextEditor;"))
 
 
+class DBeaver(umake.frameworks.baseinstaller.BaseInstaller):
+
+    def __init__(self, **kwargs):
+        super().__init__(name="DBeaver", description=_("Free multi-platform database tool for developers"),
+                         only_on_archs=['i386', 'amd64'],
+                         download_page="https://dbeaver.io/download/",
```
LyzardKing: the download page needs to be the api github releases page: "https://api.github.com/repos/dbeaver/dbeaver/releases/latest"

umake/frameworks/ide.py
```diff
@@ -857,6 +857,41 @@ def post_install(self):
...
+                         desktop_filename="dbeaver.desktop",
+                         required_files_path=["dbeaver"],
+                         dir_to_decompress_in_tarball="dbeaver_*",
+                         **kwargs)
+
+    arch_trans = {
+        "amd64": "x64",
```
LyzardKing: the architectures are x86_64 and x86 respectively (the ones you can see in the webpage in the previous comment

umake/frameworks/ide.py
```diff
@@ -857,6 +857,41 @@ def post_install(self):
...
+        "i386": "x32"
+    }
+
+    def parse_download_link(self, line, in_download):
+        """Parse DBeaver download links"""
+        url = None
+        if '.deb' in line:
```
LyzardKing: replace .deb with .tar
We don't release debs..only tarfiles

LyzardKing:
This is the atom implementation you need to copy:
https://github.com/ubuntu/ubuntu-make/blob/master/umake/frameworks/ide.py#L754

Then, you can check if it works by running:
./runtests tests/large/tests_ide.py:DBeaverTests

LyzardKing:
from the Atom class you can remove the run_for, the install_framework_parser and the if "beta"...
since you're not installing a beta version
the rest you can copy and edit.

patsevanton:
Can we install deb package?
Why install from .tar archive?

LyzardKing:
Mainly because the deb needs to be installed in specific folders which belong to root.
Where as we want to install in the user home folder, and only ask for su privileges to install dependencies

LyzardKing:
An install from deb is a different way.
It's outside of the ubuntu-make scope, which is to permit installations without special permissions (except dependencies, but those come from the repos)
Also it manages to keep the files in specific locations, so it's easier to track them in case of updates(reinstalls)/removals
From a users standpoint the experience should be exactly the same.

patsevanton:
Hou upgrade application installed from umake?

LyzardKing:
It depends, some applications (ie the intellij ides, firefox) auto-update.
The others for now can simply be reinstalled. They will pull the latest version.
There is a PR with work on a --update command, but that's what it does anyway

patsevanton:
Pull new change

LyzardKing:
You needed to remove the beta part, since dbeaver has no beta release.
Also the test fails since there is no DBeaver- folder in the tarfile.
I'll make the changes then push to this branch.

LyzardKing:
If you now run the large test it works.
It downloads the tar, installs, checks the exec, icon, launches, then removes.
The part still missing now is the medium test.
That is basically a mock of the website (github) in a docker image to speed up the tests

LyzardKing:
It still needs some work...I left a style test :/
But if you have a look at it and have any questions let me know!

LyzardKing:
@patsevanton 
Does DBeaver have no runtime dependencies?
I noticed that it's a java program...
Does it bundle a jvm? Does it need a jdk?

patsevanton:
Yes, DBeaver requires Java (JRE) 1.8+ to run.

LyzardKing:
ok, so you need to add the requirement in the init. like this: https://github.com/ubuntu/ubuntu-make/blob/master/umake/frameworks/ide.py#L868

Check what versions of the jre are compatible and add them with a | (OR).

If that's the only runtime dependency, and you can run the medium and large tests we can merge

patsevanton:
Added

LyzardKing:
Did it not simply need the jre?

patsevanton:
OMG. Changed

LyzardKing:
Added the dependency in the medium test.
Tests now work.
Merging.
Thanks!

patsevanton:
Sorry for the long answer. I have 13 of June a son was born.

LyzardKing:
Congratulations!
No worries, we got this working in the end :)

patsevanton:
Thanks!

LyzardKing:
Thank you for contributing to ubuntu-make, and for the patience!

PR #556, opened 2018-05-31
Add gnome launcher support

LyzardKing:
This adds the installed frameworks to the gnome launcher (dock)
It's a clone of the unity version, with the necessary tweaks and a simple test.


LyzardKing:
Removing this since nobody asked for it.
It was just an experiment..
The new gnome releases don't behave this way if installing from software center.

PR #560, opened 2018-06-07
Snap test branch

LyzardKing:
Test snap fixes.
The latest commit is the updated fix.
Once it works it'll be rebased properly and only the used commit will be added to the main branch.

LATEST: use core18. Add libraries suggested by snapcraft. Will not work at the moment on build.snapcraft.io


LyzardKing:
Core18 is now supported in build.snapcraft.io
Pushing this PR so the snap can work on all the snap supported ubuntu distros.

PR #561, opened 2018-06-12
Added support for Jetbrain's Toolbox

elichai:
Jetbrain's Toolbox doesn't recognize products that doesn't have `StartupWMClass` in the `.desktop` file.

### DIFF
--- a/umake/frameworks/ide.py
+++ b/umake/frameworks/ide.py
@@ -317,13 +317,16 @@ class BaseJetBrains(umake.frameworks.baseinstaller.BaseInstaller, metaclass=ABCM
         icon_path = join(self.install_path, 'bin', self.icon_filename)
         comment = self.description + " (UDTC)"
         categories = "Development;IDE;"
+        wmclass = "StartupWMClass=jetbrains-" + self.name.split(' ', 1)[0].lower()
         create_launcher(self.desktop_filename,
                         get_application_desktop_file(name=self.name,
                                                      icon_path=icon_path,
                                                      try_exec=self.exec_path,
                                                      exec=self.exec_link_name,
                                                      comment=comment,
-                                                     categories=categories))
+                                                     categories=categories,
+                                                     extra=wmclass,
+                                                     ))
 
     def install_framework_parser(self, parser):
         this_framework_parser = super().install_framework_parser(parser)


LyzardKing:
This can't work with the name.. it might work with the desktop-filename as is...
So for instance pycharm CE has WMClass jetbrains-pycharm-ce.
And that is exactly the desktop filename.
I'm not sure of the other ones thought...

elichai:
With some quick testing I've found out that Pycharm-Professional is `jetbrains-pycharm`, Idea-Ultimate is `jetbrains-idea`, and Idea-CE(Community Edition) is `jetbrains-idea-ce`.

It does sound like the desktop-filename is a better variable to use, I'll change this.

Another thing, how do you know what the WMClass of Pycharm CE? (there's no `Tools > Create Desktop Entry` there, and there's no `.desktop` file in the gzip)

About the tests I'll need to study your testing methods to see how easy it will be for me to implement this, as I said I'm not sure where in the gzip I should find the `WMClass` variable. (I tried grepping the whole package with no luck)

LyzardKing:
If you launch it, in the splash screen (where you can select new project, open, ...)
in the bottom row there's an option for creating the desktop entry.
I don't think there's a .desktop file in the gzip.
could we find it in the toolbox? (I have never used it, so I have no idea..)
If I were to guess, I'd say they're using the desktop filename..but I can't possibly be sure.

The best thing would be to find something in the website or even better in the tarfile...

elichai:
Ok.

Right now what I tried is extracting: `idea-IC,idea-IU,jebtrains-toolbox,pycharm,pycharm-edu` into a directory and used `grep -rnw . -e` to search for: `"*WMClass*", "WMClass", "jetbrains-"` with no luck.

LyzardKing:
I worry that it might be easier to install them all ;)
Or, you could write on the intellij forum and ask some higher up for some insight,
they might be helpful!

elichai:
That's the answer I got from their forum:

> Not sure.
> Here is the code for the reference: https://github.com/JetBrains/intellij-community/blob/19f38b2462cf3d26bbad7e0615ab5b0e104c5f7c/platform/platform-impl/src/com/intellij/ide/actions/CreateDesktopEntryAction.java#L148

LyzardKing:
That's a start!
It get's us to the function where they build the WMClass:
https://github.com/JetBrains/intellij-community/blob/master/platform/platform-impl/src/com/intellij/ui/AppUIUtil.java#L165

elichai:
Yeah,
So according to that function every WMClass starts with the `VENDOR_PREFIX` which is: `private static final String VENDOR_PREFIX = "jetbrains-";`
and then they add the "short" name of the app (`.replace("intellij-idea", "idea").replace("android-studio", "studio")`)
and after that if it's a community edition they add `-ce` but if it's a professional or ultimate edition they add nothing:
`.replace("-community-edition", "-ce").replace("-ultimate-edition", "").replace("-professional-edition", "");`

PR #564, opened 2018-08-06
Remove overrides and parse json

LyzardKing:
This PR replaces #559 
This adds a json parser directly in baseinstaller, reducing the number of frameworks that need to override get_metadata_and_check_license
It's a proof of concept, but mostly working.
Also If we need to read a file for the checksum we don't need to override everything
It adds a json property to the frameworks that need it
This simplifies the readability, especially for newcomers. 
And also simplifies maintenance, if the main codebase is in one place.


LyzardKing:
@redmcg  It works!!
I'll now check the non eclipse tests, to verify that everything is working with this repo.
If so we can merge!
Thank you for the PR and the help.
If you can test something let me know.

redmcg:
Just as an FYI - it looks like it is still using HTTP instead of HTTPS for retrieving the checksum. 

You can change the replace line that's already there to get the checksum in HTTPS with:
`self.new_download_url = p.group(1).replace('download.php', 'sums.php').replace('http://', 'https://')`

And if you are worried about not all mirrors having HTTPS - then you can still do the archive download with HTTP by changing the line in `get_sha_and_start_download` to:
`url = self.new_download_url.replace('sums.php', 'download.php').replace('https://', 'http://') + '&r=1'`

LyzardKing:
it's a bit ugly with all the replace methods..but I think it's something useful to do.
I'm now running all the medium tests... then it's the turn of the large tests

TomaszDom:
Did the test pass?

LyzardKing:
Small and medium tests work.
The large tests in a vm still don't work (an unrelated issue) so I tested a few frameworks myself.
Everything seems to be working.
I'll clean the PR and merge. Hopefully it'll still work ;)

LyzardKing:
Merging.
Fixes #572 #565

mchtower:
Should the issue in 572 be resolved? I'm still getting the same error.

LyzardKing:
Are you running the git version?
The ppa takes some time to update.
On my system I can install eclipse without problems

mchtower:
I'm using ppa, not sure exactly what you mean by git version

LyzardKing:
The ppa updated 4 hours ago. Try updating.
The version if you run `apt policy ubuntu-make` should be:
> 18.05+201809191700~ubuntu18.04.1

with your ubuntu version at the end.

mchtower:
Mine reads 

18.05+2018070313530~ubuntu17.10.1

which is confusing because I have ubuntu18.04 installed

LyzardKing:
There's an issue with the PPA you added.
It's pointing to the wrong release and is not updated.
Try adding it again if possible.

mchtower:
Deleting the existing ppa and adding it back worked. 

Thanks  for all of your help

PR #577, opened 2018-11-11
eclipse downloads for some reason removed the http prefix from the do…

pellcorp:
…wnload urls

### DIFF
--- a/umake/frameworks/ide.py
+++ b/umake/frameworks/ide.py
@@ -93,9 +93,10 @@ class BaseEclipse(umake.frameworks.baseinstaller.BaseInstaller, metaclass=ABCMet
         else:
             in_download = False
         if in_download:
-            p = re.search(r"href='(http://www\.eclipse\.org\/downloads/download\.php\?file=.*\.tar\.gz)'", line)
+            p = re.search(r"href='(http:|https:)?(//www\.eclipse\.org\/downloads/download\.php\?file=.*\.tar\.gz)'",
+                          line)
             with suppress(AttributeError):
-                self.new_download_url = p.group(1).replace('download.php', 'sums.php').replace('http://', 'https://')
+                self.new_download_url = 'https:' + p.group(2).replace('download.php', 'sums.php')
                 self.https = True if parse.splittype(self.new_download_url)[0] is "https" else False
         return ((None, None), in_download)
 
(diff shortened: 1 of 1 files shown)


umake/frameworks/ide.py
```diff
@@ -93,9 +93,9 @@ def parse_download_link(self, line, in_download):
         else:
             in_download = False
         if in_download:
-            p = re.search(r"href='(http://www\.eclipse\.org\/downloads/download\.php\?file=.*\.tar\.gz)'", line)
+            p = re.search(r"href='(//www\.eclipse\.org\/downloads/download\.php\?file=.*\.tar\.gz)'", line)
```
LyzardKing:
```suggestion
            p = re.search(r"href='(http:|https:)?(//www\.eclipse\.org\/downloads/download\.php\?file=.*\.tar\.gz)'", line)
```

I'd change the regex so it works with and without the https prefix.
The checksum pages seem to all be available via https, so we don't need to check that.

umake/frameworks/ide.py
```diff
@@ -93,9 +93,9 @@ def parse_download_link(self, line, in_download):
...
             with suppress(AttributeError):
-                self.new_download_url = p.group(1).replace('download.php', 'sums.php').replace('http://', 'https://')
+                self.new_download_url = 'https:' + p.group(1).replace('download.php', 'sums.php').replace('http://', 'https://')
                 self.https = True if parse.splittype(self.new_download_url)[0] is "https" else False
```
LyzardKing: This line can be removed
The lines...107-108 with the if can be removed

umake/frameworks/ide.py
```diff
@@ -93,9 +93,9 @@ def parse_download_link(self, line, in_download):
         else:
             in_download = False
         if in_download:
-            p = re.search(r"href='(http://www\.eclipse\.org\/downloads/download\.php\?file=.*\.tar\.gz)'", line)
+            p = re.search(r"href='(//www\.eclipse\.org\/downloads/download\.php\?file=.*\.tar\.gz)'", line)
             with suppress(AttributeError):
-                self.new_download_url = p.group(1).replace('download.php', 'sums.php').replace('http://', 'https://')
+                self.new_download_url = 'https:' + p.group(1).replace('download.php', 'sums.php').replace('http://', 'https://')
```
LyzardKing:
```suggestion
                self.new_download_url = 'https:' + p.group(2).replace('download.php', 'sums.php')
```

pellcorp:
I am working on the changes

pellcorp:
still fails due to line length issues, which is weird because when I open the file its not 128 characters long.  Anyway I don't have a dev env setup for ubuntu-mate but when I have some time I will so I can try and get the PR merged.

LyzardKing:
With the changes i suggested it seems to work even in the tests, so let me know if there's something you need to verify

LyzardKing:
The line too long issue is because you need to split it like this:
```
            p = re.search(r"href='(http:|https:)?(//www\.eclipse\.org\/downloads/download\.php\?file=.*\.tar\.gz)'",
                          line)
```

pellcorp:
yay

LyzardKing:
It looks good!
I'll just run some extra tests on the other eclipse installers, but it should be fine.
Thanks!

PR #580, opened 2018-11-17
Add openjdk-11 LTS

LyzardKing:


LyzardKing:
According to upstream docs eclipse, intellij, and maven support openjdk-11.
I'm adding it to the frameworks.
Like before, the order of installation is openjkd-7, openjdk-8 and openjdk-11
If one of them is installed it will not be removed, and no other version will be installed.

PR #590, opened 2019-01-09
Fix autopkgtests on disco

didrocks:
Autopkgtests on disco failed due to new exception raised on invalid SSL.
Keep backward compatibility for snap built on older version.

I just uploaded the fix to disco FYI (without making a new release): https://launchpad.net/ubuntu/+source/ubuntu-make/18.09+disco1


LyzardKing:
Great! I've ignored the disco integration, but it needed to be checked! :)
I'll make a new release before the disco feature freeze, to add the latest fixes.
I'll notify you when I do!
Merging now!

didrocks:
Thanks a lot! :)

PR #607, opened 2019-07-22
Add terraform support

### LINKED ISSUE #605: Add terraform support
`terraform` is very popular tool nowadays for most of the backend developers and devops.
And it is updated quite often making us do a lot of extra work to update it.
So it would be really nice to have `terraform` support in `umake`.

aleks-sidorenko:
* Terraform support is added
* New category `devops` is added
* Tests are written
* Closes #605


aleks-sidorenko:
> It looks good.
> If you can simply remove the logger instructions from the parse_download_link method..
> They aren't really needed since the logs are in the baseInstaller method, so are applied to all the frameworks.

This is done

PR #617, opened 2020-02-22
Update SpringToolsSuite download_page

leonavam:
current rout: "https://spring.io/tools/"


LyzardKing:
Thanks for the fix!
The tests would need changing.. let me know if you can look at those..

PR #639, opened 2020-09-03
List depends

LyzardKing:
Add the functionality to list the dependencies for a framework without installing.
The tests are working.
If a framework is not installable then it'll not be shown (this behaviour can be changed)


rellyrell7898:
# ~~****~~

PR #645, opened 2020-09-30
Update README.md

HexaOneOfficial:


LyzardKing:
This change is not in line with the recommended way to use badges, as described in https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/adding-a-workflow-status-badge

PR #660, opened 2021-10-02
sdVersion update

ronytan:
sdsdsds


LyzardKing:
This is just my code. It has not yet been merged because it breaks some tests.
If you want to try and fix it and work on adding other update parsers it would be helpful, but at the current state it is not ready for merging.

PR #693, opened 2023-10-16
add --update option

NicolasKeita:
add --update option (or -u) to enable comparing the user's version of each framework to the latest version available.

Inspired by an old issue #74 and I wanted to help.

Side nodes:
- This new feature does not fully automate the update of **all** outdated packages (it proposes to the user to only update **one** package) but presents a clear indication of which frameworks are outdated (there are rooms for improvements).
- I did not create test-cases for this feature but I didn't break any previous successful tests neither. (There are rooms for improvements).

Feature details:
- I get the framework's latest version from parsing the package **url**.
- I get the current user version of the framework by reading files in the install_path (most of the times, there is a version.txt file. I handle it differently for each framework).
- Then I compare and display which framework is outdated in the terminal as shown in the image below.

image

### DIFF
--- a/umake/frameworks/__init__.py
+++ b/umake/frameworks/__init__.py
@@ -30,6 +30,7 @@ import os
 import pkgutil
 import sys
 import subprocess
+import re
 from umake.network.requirements_handler import RequirementsHandler
 from umake.settings import DEFAULT_INSTALL_TOOLS_PATH, UMAKE_FRAMEWORKS_ENVIRON_VARIABLE, DEFAULT_BINARY_LINK_PATH
 from umake.tools import ConfigHandler, NoneDict, classproperty, get_current_arch, get_current_distro_version,\
@@ -140,7 +141,7 @@ class BaseFramework(metaclass=abc.ABCMeta):
     def __init__(self, name, description, category, force_loading=False, logo_path=None, is_category_default=False,
                  install_path_dir=None, only_on_archs=None, only_ubuntu=False, only_ubuntu_version=None,
                  packages_requirements=None, only_for_removal=False, expect_license=False,
-                 need_root_access=False, json=False, override_install_path=None):
+                 need_root_access=False, json=False, override_install_path=None, version_regex=None):
         self.name = name
         self.description = description
         self.logo_path = None
@@ -153,6 +154,8 @@ class BaseFramework(metaclass=abc.ABCMeta):
         self.packages_requirements.extend(self.category.packages_requirements)
         self.only_for_removal = only_for_removal
         self.expect_license = expect_license
+        self.version_regex = version_regex
+        self.forbidden_to_update = False
         # self.override_install_path = "" if override_install_path is None else override_install_path
 
         # don't detect anything for completion mode (as we need to be quick), so avoid opening apt cache and detect
@@ -330,6 +333,14 @@ class BaseFramework(metaclass=abc.ABCMeta):
             self.setup(install_path=install_path,
                        auto_accept_license=auto_accept_license,
                        dry_run=dry_run)
+
+    def get_latest_version(self):
+        return (re.search(self.version_regex, self.package_url).group(1).replace('_', '.')
+            if self.package_url and self.version_regex else None)
+
+    @staticmethod
+    def get_current_user_version(install_path):
+        return None
 
 
 class MainCategory(BaseCategory):
--- a/umake/frameworks/__init__.py
+++ b/umake/frameworks/__init__.py
@@ -141,7 +141,8 @@ class BaseFramework(metaclass=abc.ABCMeta):
     def __init__(self, name, description, category, force_loading=False, logo_path=None, is_category_default=False,
                  install_path_dir=None, only_on_archs=None, only_ubuntu=False, only_ubuntu_version=None,
                  packages_requirements=None, only_for_removal=False, expect_license=False,
-                 need_root_access=False, json=False, override_install_path=None, version_regex=None):
+                 need_root_access=False, json=False, override_install_path=None,
+                 version_regex=None, supports_update=False):
         self.name = name
         self.description = description
         self.logo_path = None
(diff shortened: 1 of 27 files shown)


LyzardKing:
Hi. Thanks for contributing this.
The update functionality has been on my mind for a while.
Unfortunately I don't have too much time to check this, but I'll get to it soon.

Just a first few suggestions:
- I would avoid printing everything, and instead only show the updatable items.
- keep the default "Missing Information" functions in baseInstaller, and just override if needed.

In a previous attempt I started a while back I added an argument (updatable) to distinguish the frameworks that shouldn't be updatet through umake (e.g. those that have their own updater).

Finally, more an idea than anything else, but I think it could be better to not read the version from the url, since it may not be there (think of a generic url that always points to the latest version), but instead save a config file that logs the installed version on install.

LyzardKing:
You can find the old version of the update function here:
https://github.com/ubuntu/ubuntu-make/pull/506

Just to have a quick look at the defaults and "updatable" argument.

LyzardKing:
Hi @NicolasKeita 
I haven't had much time to look into this, but I'm updating a few things (packaging mainly) on the `new_snap` branch.
I need to simplify/remove most of the tests, since they are not maintained.

Did you manage to look at making the update function more modular?

NicolasKeita:
Hi,

I believe that I approached the task with the mindset of adding my code on top of the existing code rather than modifying it to avoid any potential disruptions.
If the tests could be revised, I would be more confident in improving my pull request.
I believe I can make it more modular :)

LyzardKing:
Thank you very much for the patience and willingness to look into this.

My proposal would be to split this in two: 
- add a generic "default" method in frameworks/__init__.py and frameworks/baseInstaller.py (as in the previous attempt I linked to)
- override that method only in frameworks that can update via umake (things like intellij should not because they have their own update system)

- add a --list-upgradable in __init__.py that shows what you did now, so a list of upgradable frameworks.. I would not format them  as a table, or anything fancy, trying instead to reuse most of the list method that exists now.

Let me know what you think, and thank you again!

NicolasKeita:
Thanks.

I understood the proposal. Looks good to me. I'll take your pull request (#506) code as base. I should be submitting something in the next few days.

LyzardKing:
Keep in mind that my old pull request is very old now, so it's not up to date. Just look at it for pointers, or rebase that on master. The first one might be easier.

Let me know if you need help with anything!

The changes I'll be making should not interfere with whatever you do, and at the moment they live in different branches.

LyzardKing (commented):
I gave your code another look,
I think it's great, and the only thing required is to avoid repeating too much code in the single frameworks.

If possible keep the default methods (those that say missing information) in baseinstaller, and the frameworks only override them if they are updatable by umake.

umake/frameworks/__init__.py
```diff
@@ -331,6 +334,14 @@ def run_for(self, args):
...
                        dry_run=dry_run)
 
+    def get_latest_version(self):
+        return (re.search(self.version_regex, self.package_url).group(1).replace('_', '.')
+            if self.package_url and self.version_regex else None)
+
+    @staticmethod
+    def get_current_user_version(install_path):
+        return None
+
```
NicolasKeita, line 344: Adding two default methods for each framework.
get_latest_version() is parsing the download url string to extract the version.
get_current_user_version() is overridden in each of the framework. It is checking the installed files to fetch the version.

NicolasKeita, line 344: Adding two default methods for each framework.
get_latest_version() is parsing the download url string to extract the version.
get_current_user_version() is overridden in each of the framework. It is checking the installed files to fetch the version.

NicolasKeita:
the pretty_print function can be easily remove if needed. I filtered only the outdated frameworks (compared to printing every framework installed)

Screenshot from 2024-01-10 05-18-05

umake/frameworks/__init__.py
```diff
@@ -153,6 +154,8 @@ def __init__(self, name, description, category, force_loading=False, logo_path=N
         self.packages_requirements.extend(self.category.packages_requirements)
         self.only_for_removal = only_for_removal
         self.expect_license = expect_license
+        self.version_regex = version_regex
+        self.forbidden_to_update = False
```
NicolasKeita, line 158: This was my way of preventing an update on some frameworks like Intellij.
Set by default at False then overridden in the class BaseJetBrains for example.

umake/ui/cli/__init__.py
```diff
@@ -322,30 +315,34 @@ def main(parser):
...
-            run_command_for_args(args)
-
-            return
+            pretty_print_versions(outdated_frameworks)
+            for outdated_framework in outdated_frameworks:
+                if outdated_framework['forbidden_to_update'] is False:
+                    args = parser.parse_args([outdated_framework['category_name'], outdated_framework['framework_name']])
+                    CliUI()
+                    run_command_for_args(args)
+                    return
```
NicolasKeita, line 340: If multiple outdated frameworks are present, only the first framework updates (the return statement is early).
I struggled a bit with running CliUI() in a loop.

umake/frameworks/__init__.py
LyzardKing: I would have used it the other way around, keeping the default as not updatable, and setting a framework to updatable explicitly. This way we only add information for those that have this functionality.

What do you think?

umake/ui/cli/__init__.py
LyzardKing: I'll look into this.
The best way would probably be to relaunch umake, but I'll let you know

LyzardKing:
Great!

The only thing missing now is to add a couple of tests. Do you think you could look into that? Specifically a "small" test (you can find them in the tests/small folder), to see if the behaviour is consistent.

I've been looking into the multiple update, and I think I have an idea, but I need to test it out. Basically we should have a way to run umake with multiple frameworks first (for the install as well), then we could reuse that for the update.

NicolasKeita:
Yes, working on it :+1:

NicolasKeita:
Hi,

I couldn't add proper tests.
I've faced difficulties with simple tasks, such as initializing a framework object or installing a framework.
I've added a placeholder for tests in the meantime. I don't think I am able to perform the tests :disappointed:

LyzardKing:
I'll look at adding a few tests, just to check that future updates don't break everything, then I'll merge.
I might need some time though, since work is taking over at the moment.
I'll keep you posted and hopefully this can be merged soon.

Thanks again!

LyzardKing:
Merging in a separate branch to check and add tests if possible.
It'll be then merged in master

PR #698, opened 2024-06-13
Adapt to the changes in godot download page

Jeffrey04:
This is a minor patch to fix godot installation with umake. Long term fix would be introducing a proper html parser but for now this fix will do.


Jeffrey04:
hmm the asset logo url is changed again, let me fix it

LyzardKing (changes requested):
Looks good.
Change the regex to `href=\"?([^\s]+\.zip)` since we don't need the rest of the line parsed

Jeffrey04:
> Looks good. Change the regex to `href=\"?([^\s]+\.zip)` since we don't need the rest of the line parsed
done

LyzardKing:
Merged. Thank you for the fix

